PR #52, opened 2016-10-19
missing files

Endimion75:


local-minimum:
It looks like the baseURL change in one of the files are wrong, could you fix that?

Endimion75:
Hi,

                Not sure what you mean, could you clarify?

                Luciano

PR #123, opened 2017-05-30
Experimental

skymandr:
Fixes issues 117 and 118. 117 was two typos, 118 was due to floats used inappropriately (in one case instance of `ndarray` should have had `dtype=bool`, in the other slice limits needed safe casting to `int`).


local-minimum:
Looks good to me.

PR #127, opened 2017-06-08
Remove IP filtering

gdetrez:
IP filtering was interfering with running scan-o-matic in docker in debug mode. In addition, this is better
left to the kernel, or a separate firewall.

The second commit installs dependency earlier in the Dockerfile, making rebuilding the image faster if requirements haven't changed.


scanomatic/ui_server/ui_server.py
```diff
@@ -552,15 +541,9 @@ def _fixtures():
...
-            if debug:
+        if debug:
                 _logger.info("Running in debug mode.")
-            app.run(port=port, debug=debug)
-        else:
-            if debug:
-                _logger.warning("Debugging is only allowed on local servers")
-            else:
-                app.run(port=port, host=host)
+        app.run(port=port, host=host, debug=debug)
```
local-minimum: The `try` should probably only be around the `app.run` and not the logging anymore

PR #129, opened 2017-06-16
Dev environment

local-minimum:
This fixes lingering issues with `--local` and `--global` flags and makes it so the included settings when running `docker-compose up` is debug mode while still having the ability to deploy non-debug versions (hopefully).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,3 @@ COPY setup_tools.py /tmp/setup_tools.py
 COPY get_installed_version.py /tmp/get_installed_version.py
 
 RUN cd /tmp && python setup.py install --default
-
-CMD scan-o-matic --no-browser
-
-EXPOSE 5000
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,3 +12,4 @@ COPY get_installed_version.py /tmp/get_installed_version.py
 
 RUN cd /tmp && python setup.py install --default
 CMD scan-o-matic --no-browser
+EXPOSE 5000
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,3 +11,4 @@ COPY setup_tools.py /tmp/setup_tools.py
 COPY get_installed_version.py /tmp/get_installed_version.py
 
 RUN cd /tmp && python setup.py install --default
+CMD scan-o-matic --no-browser
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,9 +8,10 @@ COPY scripts/ /tmp/scripts/
 COPY scanomatic/ /tmp/scanomatic/
 COPY setup.py /tmp/setup.py
 COPY setup_tools.py /tmp/setup_tools.py
+COPY get_installed_version.py /tmp/get_installed_version.py
 
 RUN cd /tmp && python setup.py install --default
 
-CMD scan-o-matic --no-browser --global
+CMD scan-o-matic --no-browser
 
 EXPOSE 5000
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,0 +1,5 @@
+scanomatic:
+    ports:
+        - 5000:5000
+        - 12451:12451
+    command: scan-o-matic --no-browser --debug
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -8,9 +8,6 @@ COPY scripts/ /tmp/scripts/
 COPY scanomatic/ /tmp/scanomatic/
 COPY setup.py /tmp/setup.py
 COPY setup_tools.py /tmp/setup_tools.py
+COPY get_installed_version.py /tmp/get_installed_version.py
 
 RUN cd /tmp && python setup.py install --default
-
-CMD scan-o-matic --no-browser --global
-
-EXPOSE 5000
```
gdetrez: I think it's still useful to have a default command and expose the port so that you can run it with just `docker run`

scanomatic/io/paths.py
```diff
@@ -176,6 +174,13 @@ def __one_init__(self, *args):
         self.scan_project_file_pattern = "{0}.scan.instructions"
         self.scan_log_file_pattern = "{0}.scan.log"
 
+    @staticmethod
+    def _make_directory(path):
+
+        if os.path.isdir(path) is False:
```
gdetrez: Why not `if not os.path.isdir(path):`

scanomatic/ui_server/ui_server.py
```diff
@@ -66,19 +66,13 @@ def launch_server(is_local=None, port=None, host=None, debug=False):
...
     if port is None:
         port = Config().ui_server.port
-
-    if is_local is True or (Config().ui_server.local and is_local is None):
-        set_local_app()
-        host = "localhost"
-        is_local = True
-    elif host is None:
+    if host is None:
         host = "0.0.0.0"
```
gdetrez: Could be the default value in the function signature

Dockerfile
local-minimum: Thing is, this was running instead of/before the docker-compose settings.

scanomatic/io/paths.py
local-minimum: Good question :+1:

scanomatic/ui_server/ui_server.py
local-minimum: It would require the start-up script to better check what arguments are set, because now it override the default anyhow. So None of them should really have defaults in the `launch_server`, but I agree in that it should be different so I make it work like the port.

Dockerfile
skymandr: Better to keep the Dockerfile useful and figure out how to override CMD in the docker-compose file (it is doable, maybe with `- entrypoint=`, but I don't remember exactly).

docker-compose.override.yml
```diff
@@ -0,0 +1,5 @@
+scanomatic:
+    ports:
+        - 5000:5000
+        - 12451:12451
```
skymandr: What was the reason we exposed this one to begin with, and why don't we want to do it?

docker-compose.override.yml
```diff
@@ -0,0 +1,5 @@
...
+    command: scan-o-matic --no-browser --debug
```
skymandr: As noted, I think maybe this should be `entrypoint:` rather than `command:` in order to override the one in the Dockerfile. This is confusing, I know...

scanomatic/io/paths.py
```diff
@@ -109,6 +98,7 @@ def __one_init__(self, *args):
         self.fixture_grid_history_pattern = "{0}.grid.history"
```
skymandr: This initalisation is very messy. Probably not a lot to do about that, but it _might_ be nicer for the reader and easier to follow, if instead of `{<number>}` in the format strings, we used `{<name/descriptor>}`. This is just a comment, not mandatory.

docker-compose.override.yml
local-minimum: the second one is for the rpc-server and that is nice so you can interact with it in ipython directly but in production there's no reason why it should be exposed. I agree moving 5000 to the docker-file makes more sense.

scanomatic/io/paths.py
local-minimum: In general I agree but perhaps not so much so when you only have one of them. It also becomes more messy remembering what you called everything in this case where the paths are all centralized (which is a topic for another discussion).

Issue #117, opened 2017-04-24
API: compress colony response has typo on returned parameters


Endimion75:
When using the  API path:

/api/data/ccc<ccc_identifier>/image/<image_identifier>/plate/<plate_number>/compress/colony/<x>/<y>

the expected response has a breaking typo:

"sucecss" instead of success

Endimion75:
The typo seems to be related when the compression is successful only

skymandr:
Resolved by https://github.com/local-minimum/scanomatic/commit/47de2facdb118378f03d0459deafb7d98b8657b6

Issue #116, opened 2017-04-19
Flask response with NaN 


local-minimum:
It seems that having `numpy.nan` in the returned json data produces an error response with status `200/OK` while still delivering the data.

Suggesting replacing all `nan` and `inf` with `None` or similar.

local-minimum:
The real problem might be that no abort codes were ever used as they should have been. I've started rewriting api to return correct status-codes.

local-minimum:
@skymandr pull request #135 converts some more. It would be good to go through the endpoints and fix this for all of them, but there's a risk that the web-app will malfunciton in many places. @Endimion75 do you think the app will handle other status-codes or will there be an extensive need of testing all the QC-features?

Endimion75:
@local-minimum: I am not sure what are you referring to, but the UI is hardwire to expect certain values from the API if you change them, the UI will break naturally.

skymandr:
Close?

PR #130, opened 2017-06-16
Abort responses

local-minimum:
This is a partial fix to the problem of how the api gave status-code 200 on gracefully refusing or failing requests. It will now give reasonable status-code for the example end-point. Then all other endpoints need updating once agreed that this is nice way.

### DIFF
--- a/scanomatic/ui_server/calibration_api.py
+++ b/scanomatic/ui_server/calibration_api.py
@@ -22,7 +22,7 @@ from scanomatic.data_processing import calibration
 from scanomatic.data_processing.calibration import (
     add_calibration, calculate_polynomial,
     load_calibration, validate_polynomial, CalibrationValidation,
-    remove_calibration, get_data_file_path,
+    delete_ccc, get_data_file_path,
 )
 from .general import (
     serve_numpy_as_image, get_grayscale_is_valid, valid_array_dimensions,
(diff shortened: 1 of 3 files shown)


scanomatic/ui_server/calibration_api.py
```diff
@@ -463,63 +466,89 @@ def calibration_compress(ccc_identifier, image_identifier, plate, x, y):
 
         image_data = calibration.get_image_json_from_ccc(ccc_identifier, image_identifier)
         if image_data is None:
-            return jsonify(success=False, is_endpoint=True, reason="The image or CCC don't exist")
+            return json_abort(
```
skymandr: We had issues with these jsonresponses before (see issue #117). Tests would be nice.

local-minimum: Yes, there's also #116 and might be yet another issue relating to this.

local-minimum: As for tests, that would require system tests at the moment, but yes, it would be nice though the whole thing is really quite messy. Or if you just meant the `json_abort` function in itself?

skymandr: I agree that system tests would be better, but are perhaps out of scope for this pull request, but yes, that was what I meant. A test of `json_abort` function seems trivial, but yes, please add that. If nothing else, then you will have created a file for putting the tests in...

local-minimum: Wow that was a bit more tricky than expected but now I have two tests.

skymandr: Good work!

PR #131, opened 2017-06-19
CCC data upload

local-minimum:
This implements endpoint for uploading data, stores copy and parses it into the CCC-json file.
I have some tests on it but setting it up so that a reasonable test of actually uploading the file, it is quite a lot of work.


scanomatic/io/meta_data.py
```diff
@@ -261,6 +261,24 @@ def __setstate__(self, state):
...
+
+            plate = []
+            plates.append(plate)
+
+            for id_outer in range(outers):
+
+                data = []
+                plate.append(data)
+
+                for id_inner in range(inners):
```
skymandr: Three nested for loops? This is not like you... :)

scanomatic/data_processing/calibration.py
```diff
@@ -971,3 +976,75 @@ def remove_calibration(label, degree=None, file_path=None):
...
+                plates.append(plate[CCCPlate.grid_shape])
+                cells.append(plate[CCCPlate.compressed_ccc_data])
+
+    return plates, cells
+
+
+@_validate_ccc_edit_request
+def add_external_data_to_ccc(identifier, data_file, report):
+
+    report['warnings'] = []
```
skymandr: Can we assume that there are no warnings already present at this stage now and for the foreseeable future? I.e. are there, and will there not be, any steps preceding this that may want to log errors and warnings? I'm asking since report is passed as an argument.

scanomatic/data_processing/calibration.py
```diff
@@ -971,3 +976,75 @@ def remove_calibration(label, degree=None, file_path=None):
...
+
+    grid_shapes, measurements = _get_all_grid_shapes(ccc)
+    meta_data = MetaData(grid_shapes, file_path)
+    if not meta_data.loaded:
+        report['errors'] = (
+            'File could not be understood in terms of plates included'
+            '(Grid shapes: {0})'.format(grid_shapes))
+        return False
+
+    report['errors'] = []
```
skymandr: Se above.

scanomatic/io/meta_data.py
local-minimum: nope but I figured clarity for once

skymandr: I approve! ;)

scanomatic/data_processing/calibration.py
local-minimum: Currently yes, but you are right that it's a bit of a naughty side-effect to wipe previous state if it exists. I'll fix

skymandr: OK. Bear in mind when fixing, however, that there is a check bellow on whether errors is empty.

PR #132, opened 2017-06-21
CORS in debug mode

local-minimum:
Luciano requested that we allow CORS while running in debug mode. This opens up for http://en.wikipedia.org/wiki/Cross-site_request_forgery but given that it is only active in debug I think it should be OK.

### DIFF
--- a/scanomatic/ui_server/ui_server.py
+++ b/scanomatic/ui_server/ui_server.py
@@ -537,13 +537,23 @@ def launch_server(host, port, debug):
 
     if debug:
         CORS(app)
-        _logger.info("Running in debug mode, allowing CORS.")
+        _logger.warning(
+            "\nRunning in debug mode, causes sequrity vunerabilities:\n" +
+            " * Remote code execution\n" +
+            " * Cross-site request forgery\n" +
+            "   (https://en.wikipedia.org/wiki/Cross-site_request_forgery)\n" +
+            "\nAnd possibly more issues"
+
+        )
     try:
         app.run(port=port, host=host, debug=debug)
     except error:
-        _logger.warning("Could not bind socket, probably server is already running and this is nothing to worry about."
-                        "\n\tIf old server is not responding, try killing its process."
-                        "\n\tIf something else is blocking the port, try setting another port using --help.")
+        _logger.warning(
+            "Could not bind socket, probably server is already running and" +
+            " this is nothing to worry about." +
+            "\n\tIf old server is not responding, try killing its process." +
+            "\n\tIf something else is blocking the port," +
+            " try setting another port using --help.")
         return False
     return True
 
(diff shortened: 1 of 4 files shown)


Endimion75:
I tested Martin's CORS changes and they work as expected.  This feature is very useful when testing the API when the API and the UI are in the same machine.

scanomatic/ui_server/ui_server.py
```diff
@@ -535,7 +536,8 @@ def _fixtures():
     calibration_api.add_routes(app)
 
     if debug:
-        _logger.info("Running in debug mode.")
+        CORS(app)
+        _logger.info("Running in debug mode, allowing CORS.")
```
skymandr: I think "allowing CORS" is not very informative. Either write it as a warning, so that the user understands that this is something they should care about (CORS is not an immediately recognisable acronym) or write more verbosely what it means. Or write a general warning, pointing the user to somewhere where they can find clear information about what debug mode activates. The last option might be the most future safe, though it means another resource that has to be maintained. Oh the bother...

Endimion75:
What if you shift the burden of knowing what CORS is to the user, by using a --CORS flag to activate, rather than being a default on the --debug flag.  Nothing to maintain then.

scanomatic/ui_server/ui_server.py
Endimion75: What if you shift the burden of knowing what CORS is to the user, by using a --CORS flag to activate, rather than being a default on the --debug flag. Nothing to maintain then.

skymandr:
Hmm... Yes, that seems both safer and perhaps easier for everyone, though we'll have to watch out for feature creep.

scanomatic/ui_server/ui_server.py
local-minimum: I'll go with better info (as a warning) rather than do a separate setting/argument to the app especially since CORS isn't a great issue for the app at the moment since we trust all users equally and have no authentication.

skymandr: OK!

scanomatic/ui_server/ui_server.py
```diff
@@ -537,13 +537,23 @@ def _fixtures():
...
     except error:
-        _logger.warning("Could not bind socket, probably server is already running and this is nothing to worry about."
-                        "\n\tIf old server is not responding, try killing its process."
-                        "\n\tIf something else is blocking the port, try setting another port using --help.")
+        _logger.warning(
+            "Could not bind socket, probably server is already running and" +
+            " this is nothing to worry about." +
+            "\n\tIf old server is not responding, try killing its process." +
+            "\n\tIf something else is blocking the port," +
+            " try setting another port using --help.")
```
skymandr: technically it is not "using --help", is it?

Issue #134, opened 2017-06-29
API: Request compression error response


Endimion75:
When I make a request to the API for a colony compression:

`/api/data/ccc<ccc_identifier>/image/<image_identifier>/plate/<plate_number>/compress/colony/<x>/<y>`

If I send colony data that is "wrong" I.E. blob values, neither values, but no background values.  Instead of getting a json response like so:

{
  "is_endpoint": false, 
  "reason": "Background must be consisting of at least 3 pixels", 
  "success": false
}

I get a BAD REQUEST error (status: 400) with a response object with the values I was expecting:

{
  "is_endpoint": false, 
  "reason": "Background must be consisting of at least 3 pixels", 
  "success": false
}

Should I be getting directly a success response, but a success = false json object, rather than an error response?

skymandr:
This may also have been fixed in the new pull request. Could you check if the problem persists once it is up?

Endimion75:
will do

skymandr:
There, reset the master branch and made a proper pull request. Please update your code accordingly (currently no changes, but I will commit updates to any other issues to the new branch instead - I have also assigned you as reviewer(s) so please have a look at the change to see that I haven't done anything atrocious.)

skymandr:
This behaviour is as it should be. We will successively make sure that all failed calls return something else than 200 with a description of the error. What may change is that we will likely move away from 400 across the board to more informative 40X status codes. Closing this issue.

Endimion75:
I have problems with this.
I have been constructing the CCC UI with the expectation that the API will behave in certain way.  Suddenly changing the way the API responds without consideration for backwards compatibility does not sparkle with me.

Endimion75:
I understand to return an error if the data sent was misconstrued, incorrect or missing.  But the data is correct, the API could not accommodate properly the grid, but still returns values and a fails state.  

In the past this cases were treated as successful calls with an success = false state.  And ALL of the API calls I make expect that type of behaviour. 

Now are you telling me that all that is to be overhauled?

Endimion75:
Also, if unsuccessful requests are going to be treated as error responses, what is the point of having a success = true/false field in the response?

skymandr:
Ok, so this was definitely a change that should have been a separate pull request, not included in #135... Lesson learned for the future! ;)

To me it seems there are two ways forward:
[...]

PR #135, opened 2017-06-30
Issue 133 set gridding outputs confusing

skymandr:
We currently know that this has issues, but I am creating a pull request nonetheless to get the proper review process started. Will look at the reported problem and see if I can fix it, but meanwhile would appreciate if someone would look at the changes not directly related to the grid API.

### DIFF
--- a/scanomatic/image_analysis/grid_array.py
+++ b/scanomatic/image_analysis/grid_array.py
@@ -237,7 +237,7 @@ class GridArray(object):
     @property
     def grid(self):
 
-        return self._grid.tolist()
+        return self._grid[:, ::-1, :].tolist()
 
     @property
     def grid_shape(self):
--- a/scanomatic/image_analysis/grid_array.py
+++ b/scanomatic/image_analysis/grid_array.py
@@ -224,7 +224,7 @@ class GridArray(object):
     def grid_shape(self):
 
         return self._grid.shape[1:]
-    
+
     @property
     def index(self):
         return self._identifier[-1]
@@ -263,7 +263,7 @@ class GridArray(object):
 
         if offset and not all(o == 0 for o in offset):
 
-            # The direction of the first axis is flipped to make offsetting more logical from user perspective
+            # The direction of the first (not zeroth) axis is flipped to make offsetting more logical from user perspective
             # this inversion must be matched by equal inversion in detect_grid
 
             o = offset[0] * -1
--- a/scanomatic/image_analysis/grid_array.py
+++ b/scanomatic/image_analysis/grid_array.py
@@ -18,7 +18,8 @@ from scanomatic.io.pickler import unpickle_with_unpickler
 import image_basics
 from scanomatic.models.analysis_model import IMAGE_ROTATIONS
 from scanomatic.image_analysis.grayscale import getGrayscale
-from scanomatic.models.factories.analysis_factories import AnalysisFeaturesFactory
+from scanomatic.models.factories.analysis_factories import \
+    AnalysisFeaturesFactory
 from scanomatic.data_processing.calibration import load_calibration
 #
 # EXCEPTIONS
@@ -28,7 +29,8 @@ class InvalidGridException(Exception):
     pass
 
 
-def _analyse_grid_cell(grid_cell, im, transpose_polynomial, image_index, semaphore=None, analysis_job_model=None):
+def _analyse_grid_cell(grid_cell, im, transpose_polynomial, image_index,
+                       semaphore=None, analysis_job_model=None):
 
     """
 
@@ -38,7 +40,8 @@ def _analyse_grid_cell(grid_cell, im, transpose_polynomial, image_index, semapho
 
     grid_cell.source = _get_image_slice(im, grid_cell).astype(np.float64)
     if grid_cell.source is None:
-        GridArray._LOGGER.error("Tried to analyse grid cell that doesn't have any area")
+        GridArray._LOGGER.error(
+            "Tried to analyse grid cell that doesn't have any area")
         if semaphore is not None:
             semaphore.release()
         return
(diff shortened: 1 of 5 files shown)


skymandr:
I would also very much appreciate input on how to write tests for the issue.

Endimion75:
I am currently cleaning up the CCC prototype so it can be used to test the gridding. hopefully I will be able to push the code soon

Endimion75:
I tested the gridding and even though it did not pop an error, the grid seems to be rotated:

sc3

is like the row are now cols and the cols are now rows

Endimion75:
Also Issue 134 remains, but now extended to the set grid API path:

/api/data/ccc/<ccc_identifier>/image/<image_identifier>/plate/<plate_number>/grid/set

Here I also get a bad request page status 400 or in other words an error response from the Ajax call to the API rather than a successful response, with a json object indicating that the gridding was unsuccessful.

Endimion75:
I though that the issue 134 was localized to the compression API call, but now that I saw it in the gridding call I realized that something much have changed in the API code that affected other API paths.  the CCC UI was wired to expect a behaviour that is no more.

It is troubling as belabour as fundamental as this is changing ...

Endimion75:
Regarding issue 134.

When I use the old debug_CORS branch (which is the latest stable branch I am using to continue my development) the problem with the success/error responses on the "set grid" API path doesn't exist.

So it was introduced while attempting to fix the issues in the new pull request and is not as old as I though it to be!  Which is great news!  I don't have to double check all the UI again :D

skymandr:
Regarding 134, 40X with failure description in the JSON response is the new way the API will work. We will successively phase out 200 for failed responses (most API endpoints should be converted already, but some stragglers may remain).

Regarding the core issue, we'll look into it now! =)

skymandr:
We identified the issue with the rotated grid and think we fixed it at its source in the latest commit.

local-minimum:
I'd like @Endimion75 to verify that this update now works better. We expect the grid origin / `[0, 0]` position to be in the lower left corner from now.

Endimion75:
This is what I got:

sc6

The gird is standing on the short side same as the slice, but the grid seems is still off by one row

local-minimum:
That the grid ends up in the wrong position and especially in this direction is weird, but I think that should be a separate issue to not make a too large issue out of this. I'll check the code tomorrow. And btw @Endimion75 if you could make more frequent pull requests for the UI we could check these things out ourselves unless you think this is good workflow that we ask you to test stuff, in that case everything is good.

scanomatic/image_analysis/grid_array.py
```diff
@@ -218,13 +237,13 @@ def grid_cell_size(self):
     @property
     def grid(self):
 
-        return self._grid.tolist()
+        return self._grid[:, ::-1, :].tolist()
```
local-minimum: This will sound weird but perhaps a docstring to explain the reason for inversing the y-axis

skymandr:
Added a docstring. Will work on implementing tests during the week. We need to discuss how to proceed with regard to Issue #134: What problem was `json_abort` supposed to solve, and is it worth it considering the extra work for Luciano?

skymandr:
I would also request that someone with more insight opens an issue for the problem with the inconsistent gridding results.

skymandr:
Regarding #134 and the `json_abort`s - should I revert those changes?

local-minimum:
There are already various api-calls that sends error codes in master since previous commits, and to me this will make the api and UI development much cleaner when using `promise` type  interfaces that are the coming/current standards of EC6 as well as makes the api more restful (if i get how that is supposed to work). As it is now, without sending error-codes basically all calls to the api must check for `success=False` in successful calls and check in failed calls for how to handle that the api did not agree with the call. This should mean code duplication in all api-calls. If we from the API guarantee that if the call did not go through that it will be sent back with an error code, the UI can assume that things have worked well when 200 is returned and must only handle problems in error/fail callbacks. So my vote is clearly for using abort-codes. This by all means must decrease the complexity and needs of the UI.

Endimion75:
@local-minimum:  You do what you need to do.  When done, just document somewhere the new pattern that the API will follow and I will update my end of the code.  All I ask is consistency with the new pattern across the API calls that are used by the CCC UI.  When I come back from holidays in 5 or so weeks, I will implement the new changes in the CCC UI and let you know who it went.

skymandr:
Updated with tests. The test applies a correction of `(3, 0)`, without which the gridding (when it works) looks like this:
gridding_right_correction_none
but with correction, the gridding (when it works) is correct:
gridding_right_correction_ 3 0

I write "when it works" above, since sometimes the needed offset is `(4, 0)`, but I will go into that in more detail in #138.

Conclusion is that `GridArray` works as expected, but that the gridding is sometimes wrong.

scanomatic/image_analysis/test/test_grid_array.py
```diff
@@ -0,0 +1,43 @@
...
+            grid_cell = grid_array[(row, col)]
+            try:
+                assert grid_cell.xy1[0] < grid_array.grid[0][row][col]
+                assert grid_cell.xy1[1] < grid_array.grid[1][row][col]
+                assert grid_cell.xy2[0] > grid_array.grid[0][row][col]
+                assert grid_cell.xy2[1] > grid_array.grid[1][row][col]
+            except Exception as err:
+                raise Exception("{} for cell ({}, {})".format(err, row, col))
+
+
```
local-minimum: Too many lines here

scanomatic/image_analysis/test/test_grid_array.py
```diff
@@ -0,0 +1,43 @@
...
+    image = ndimage.io.imread(
+        './scanomatic/image_analysis/test/testdata/test_fixture.tiff')
+    grid_array_instance = grid_array_module.GridArray(
+        image_identifier, pinning, analysis_model)
+    correction = (3, 0)
+    grid_array_instance.detect_grid(image, grid_correction=correction)
+    return grid_array_instance
+
+
+class TestGridDetection():
```
local-minimum: I would like a test of a grid_cell's identifier and testing that some `grid_cell`s position https://github.com/Scan-o-Matic/scanomatic/blob/master/scanomatic/image_analysis/grid_cell.py#L37 matches the position you request from the `grid_array` like with line 31 below.

PR #137, opened 2017-07-04
latest version of CCC

Endimion75:
This has the latest version of the CCC UI

I have been using this file to create the fixture and to test the UI:

[hanna_BY4741_calibration_160915_0000_33.7167.zip](https://github.com/Scan-o-Matic/scanomatic/files/1121574/hanna_BY4741_calibration_160915_0000_33.7167.zip)

This version is rigged to start gridding with the offset [4, 0] which is needed because the images used to test the CCC requires this offset.  The UI is made to detect an unsuccessful gridding and will request the user to input an offset and try again.  Unfortunately this wont work with the new breaking changes in the pull request "Issue 133 set gridding outputs confusing"

If you want to remove the rig, got to line 538 in CCC.html and uncomment the line with the offset [0,0] and comment the one with the offset [4,0]

### DIFF
--- a/data/ui_server/js/CCCHelper.js
+++ b/data/ui_server/js/CCCHelper.js
@@ -31,3 +31,169 @@ function dataURItoBlob(dataURI) {
 
     return new Blob([ia], { type: mimeString });
 }
+
+function getDataUrlfromUrl(src, callback) {
+    var img = new Image();
+    img.crossOrigin = 'Anonymous';
+    img.onload = function () {
+        var canvas = document.createElement('CANVAS');
+        var ctx = canvas.getContext('2d');
+        var dataURL;
+        canvas.height = this.height;
+        canvas.width = this.width;
+        ctx.drawImage(this, 0, 0);
+        dataURL = canvas.toDataURL();
+        callback(dataURL);
+    };
+    img.src = src;
+    if (img.complete || img.complete === undefined) {
+        img.src = "data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///ywAAAAAAQABAAACAUwAOw==";
+        img.src = src;
+    }
+}
+
+
+function createCanvasImage(data, canvas) {
+
+    var cs = getLinearMapping(data);
+    var rows = data.image.length;
+    var cols = data.image[0].length;
+    var cvPlot = canvas;
+    cvPlot.width = cols;
+    cvPlot.height = rows;
+    var ctx = cvPlot.getContext('2d');
+    var imgdata = ctx.getImageData(0, 0, cols, rows);
+    var imgdatalen = imgdata.data.length;
+    var imageCol = -1;
+    var imageRow = 0;
+    for (var i = 0; i < imgdatalen; i += 4) {  //iterate over every pixel in the canvas
+        imageCol += 1;
+        if (imageCol >= cols) {
+            imageCol = 0;
+            imageRow += 1;
+        }
+        var color = data.image[imageRow][imageCol];
+        var mappedColor = cs(color);
+
+        var rgb = hexToRgb(mappedColor);
+
+        imgdata.data[i + 0] = rgb.r;    // RED (0-255)
+        imgdata.data[i + 1] = rgb.g;    // GREEN (0-255)
+        imgdata.data[i + 2] = rgb.b;    // BLUE (0-255)
+        imgdata.data[i + 3] = 255;  // APLHA (0-255)
+
+    }
+    ctx.putImageData(imgdata, 0, 0);
+}
+
+function createCanvasMarker(data, canvas) {
+    var rows = data.image.length;
+    var cols = data.image[0].length;
+    var cvPlot = canvas;
+    cvPlot.width = cols;
+    cvPlot.height = rows;
+    var ctx = cvPlot.getContext('2d');
+    var imgdata = ctx.getImageData(0, 0, cols, rows);
+    var imgdatalen = imgdata.data.length;
+    var imageCol = -1;
+    var imageRow = 0;
+    for (var i = 0; i < imgdatalen; i += 4) {  //iterate over every pixel in the canvas
+        imageCol += 1;
+        if (imageCol >= cols) {
+            imageCol = 0;
+            imageRow += 1;
+        }
+        var rgb;
+        if (data.blob[imageRow][imageCol] === true) {
+            rgb = { r: 255, g: 0, b: 0 }
+        } else if (data.background[imageRow][imageCol] === true) {
+            rgb = { r: 0, g: 128, b: 0 }
+        } else {
+            rgb = { r: 0, g: 0, b: 0 };
+        }
+
+        imgdata.data[i + 0] = rgb.r;    // RED (0-255)
+        imgdata.data[i + 1] = rgb.g;    // GREEN (0-255)
+        imgdata.data[i + 2] = rgb.b;    // BLUE (0-255)
+        imgdata.data[i + 3] = 255;  // APLHA (0-255)
+    }
+    ctx.putImageData(imgdata, 0, 0);
+}
+
+function getMarkerData(canvasId) {
+
+    var cvPlot = document.getElementById(canvasId);
+    var rows = cvPlot.height;
+    var cols = cvPlot.width;
+
+    var ctx = cvPlot.getContext('2d');
+    var imgdata = ctx.getImageData(0, 0, cols, rows);
+    var imgdatalen = imgdata.data.length;
+    var imageCol = -1;
+    var imageRow = 0;
+
+    var blob = new Array(rows);
+    for (var i = 0; i < rows; i++) {
+        blob[i] = new Array(cols);
+    }
+
+    var background = new Array(rows);
+    for (var i = 0; i < rows; i++) {
+        background[i] = new Array(cols);
+    }
+
+    for (var i = 0; i < imgdatalen; i += 4) {  //iterate over every pixel in the canvas
+        imageCol += 1;
+        if (imageCol >= cols) {
+            imageCol = 0;
+            imageRow += 1;
+        }
+        var r = imgdata.data[i + 0];
+        var g = imgdata.data[i + 1];
+        var b = imgdata.data[i + 2];
+        var a = imgdata.data[i + 3];
+        var pixelRgbA = r + "," + g + "," + b;
+        
+        if (pixelRgbA === "255,0,0") {
+            blob[imageRow][imageCol] = true;
+            background[imageRow][imageCol] = false;
+        } else if (pixelRgbA === "0,0,0") {
+            blob[imageRow][imageCol] = false;
+            background[imageRow][imageCol] = true;
+        };
+    }
+    var x = { blob: blob, background: background };
+    return x;
+}
+
+function hexToRgb(hex) {
+    // Expand shorthand form (e.g. "03F") to full form (e.g. "0033FF")
+    var shorthandRegex = /^#?([a-f\d])([a-f\d])([a-f\d])$/i;
+    hex = hex.replace(shorthandRegex, function (m, r, g, b) {
+        return r + r + g + g + b + b;
+    });
+
+    var result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+    return result ? {
+        r: parseInt(result[1], 16),
+        g: parseInt(result[2], 16),
+        b: parseInt(result[3], 16)
+    } : null;
+}
+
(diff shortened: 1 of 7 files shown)


data/ui_server/js/CCCHelper.js
```diff
@@ -1,33 +1,199 @@
...
+function createCanvasImage(data, canvas) {
+
+    var cs = getLinearMapping(data);
+    var rows = data.image.length;
+    var cols = data.image[0].length;
+    var cvPlot = canvas;
+    cvPlot.width = cols;
+    cvPlot.height = rows;
+    var ctx = cvPlot.getContext('2d');
+    var imgdata = ctx.getImageData(0, 0, cols, rows);
```
local-minimum: Please be consistent with camelCase

data/ui_server/js/CCCHelper.js
```diff
@@ -1,33 +1,199 @@
...
+
+function getMarkerData(canvasId) {
+
+    var cvPlot = document.getElementById(canvasId);
+    var rows = cvPlot.height;
+    var cols = cvPlot.width;
+
+    var ctx = cvPlot.getContext('2d');
+    var imgdata = ctx.getImageData(0, 0, cols, rows);
+    var imgdatalen = imgdata.data.length;
```
local-minimum: camelCase

data/ui_server/js/CCCHelper.js
```diff
@@ -1,33 +1,199 @@
...
+        var r = imgdata.data[i + 0];
+        var g = imgdata.data[i + 1];
+        var b = imgdata.data[i + 2];
+        var a = imgdata.data[i + 3];
+        var pixelRgbA = r + "," + g + "," + b;
+        
+        if (pixelRgbA === "255,0,0") {
+            blob[imageRow][imageCol] = true;
+            background[imageRow][imageCol] = false;
+        } else if (pixelRgbA === "0,0,0") {
```
local-minimum: this is wrong. You encode background as `"0,128,0"`

data/ui_server/js/CCCHelper.js
```diff
@@ -1,33 +1,199 @@
...
+        imageCol += 1;
+        if (imageCol >= cols) {
+            imageCol = 0;
+            imageRow += 1;
+        }
+        var r = imgdata.data[i + 0];
+        var g = imgdata.data[i + 1];
+        var b = imgdata.data[i + 2];
+        var a = imgdata.data[i + 3];
+        var pixelRgbA = r + "," + g + "," + b;
```
local-minimum: it looks rather backwards to create a string. Given you keep your pattern you could just check the colors

```js
if (r === 255) {
//blob
} else if (b == 128) {
//background
} else {
//neither
}
```

data/ui_server/js/CCCHelper.js
```diff
@@ -1,33 +1,199 @@
...
+        r: parseInt(result[1], 16),
+        g: parseInt(result[2], 16),
+        b: parseInt(result[3], 16)
+    } : null;
+}
+
+function getLinearMapping(data) {
+
+    var colorScheme = ["white", "grey", "black"];
+    var intensityMin = data.blobMin;
```
local-minimum: why `blobMin` and not `imageMin`? Looks wrong to combine from the two.

data/ui_server/js/cccCanvasColonyMarker.js
```diff
@@ -0,0 +1,140 @@
...
+    ctx.fillStyle = this.fill;
+    ctx.beginPath();
+    ctx.arc(this.x, this.y, this.r, 0, 2 * Math.PI, false);
+    ctx.fill();
+}
+
+Blob.prototype.contains = function(mx, my) {
+    var distancesquared = (mx - this.x) * (mx - this.x) + (my - this.y) * (my - this.y);
+    return distancesquared <= this.r * this.r;
+}
```
local-minimum: I'd prefer using `Math.pow` instead

data/ui_server/js/cccCanvasColonyMarker.js
```diff
@@ -0,0 +1,140 @@
...
+
+function CanvasState(canvas) {
+    this.canvas = canvas;
+    this.width = canvas.width;
+    this.height = canvas.height;
+    this.ctx = canvas.getContext('2d');
+    var stylePaddingLeft, stylePaddingTop, styleBorderLeft, styleBorderTop;
+    if (document.defaultView && document.defaultView.getComputedStyle) {
```
local-minimum: There seems to be no handling of if this condition is not met.

data/ui_server/js/cccCanvasColonyMarker.js
```diff
@@ -0,0 +1,140 @@
...
+    this.htmlLeft = html.offsetLeft;
+
+    this.valid = false; 
+    this.shapes = [];  
+    this.dragging = false; 
+    this.selection = null;
+    this.dragoffx = 0; 
+    this.dragoffy = 0;
+  
+    var myState = this;
```
local-minimum: use `.bind(this)` instead on the `addEventListener`.

data/ui_server/js/cccCanvasColonyMarker.js
```diff
@@ -0,0 +1,140 @@
...
+    this.shapes.push(Blob);
+    this.valid = false;
+}
+
+CanvasState.prototype.clear = function() {
+    this.ctx.clearRect(0, 0, this.width, this.height);
+}
+
+CanvasState.prototype.draw = function() {
+    if (!this.valid) {
```
local-minimum: Valid sounds like the wrong word. I would prefer `toUpdate` or `needsUpdate` or `dirty`.

data/ui_server/js/cccCanvasColonyMarker.js
```diff
@@ -0,0 +1,140 @@
...
+            if (Blob.x > this.width || Blob.y > this.height ||
+                Blob.x + Blob.w < 0 || Blob.y + Blob.h < 0) continue;
+            shapes[i].draw(ctx);
+        }
+    
+        if (this.selection != null) {
+            ctx.strokeStyle = this.selectionColor;
+            ctx.lineWidth = this.selectionWidth;
+            var mySel = this.selection;
+            ctx.strokeRect(mySel.x,mySel.y,mySel.w,mySel.h);
```
local-minimum: spaces

data/ui_server/js/cccCanvasColonyMarker.js
```diff
@@ -0,0 +1,140 @@
...
+        }
+    
+        this.valid = true;
+    }
+}
+
+CanvasState.prototype.getMouse = function(e) {
+    var element = this.canvas, offsetX = 0, offsetY = 0, mx, my;
```
local-minimum: Please, one variable declaration per line.
We try to follow https://github.com/airbnb/javascript
That one is for ES6, which is not suitable for SoM yet, but principle holds:
https://github.com/airbnb/javascript#variables--one-const

data/ui_server/js/CCCHelper.js
```diff
@@ -1,33 +1,199 @@
...
+    var rows = cvPlot.height;
+    var cols = cvPlot.width;
+
+    var ctx = cvPlot.getContext('2d');
+    var imgdata = ctx.getImageData(0, 0, cols, rows);
+    var imgdatalen = imgdata.data.length;
+    var imageCol = -1;
+    var imageRow = 0;
+
+    var blob = new Array(rows);
```
local-minimum: change to:

```js
var blob = [];
for (var i = 0; i < rows; i++) {
    blob.push([]);
}
```
unless the inner dimension is specifically used to get the `cols` later in which case swap to `blob.push(new Array(cols));`

data/ui_server/js/CCCHelper.js
```diff
@@ -1,33 +1,199 @@
...
+    for (var i = 0; i < rows; i++) {
+        blob[i] = new Array(cols);
+    }
+
+    var background = new Array(rows);
```
local-minimum: same as `blob` also consider moving 2d-array creation to function

data/ui_server/js/cccCanvasColonyMarker.js
```diff
@@ -0,0 +1,140 @@
...
+}
+
+function CanvasState(canvas) {
+    this.canvas = canvas;
+    this.width = canvas.width;
+    this.height = canvas.height;
+    this.ctx = canvas.getContext('2d');
+    var stylePaddingLeft, stylePaddingTop, styleBorderLeft, styleBorderTop;
+    if (document.defaultView && document.defaultView.getComputedStyle) {
+        this.stylePaddingLeft = parseInt(document.defaultView.getComputedStyle(canvas, null)['paddingLeft'], 10)      || 0;
```
local-minimum: consider making function `getComputedCanvasStyleInt(item)` or something that just returns the int.

data/ui_server/js/CCCHelper.js
Endimion75: Corrected

data/ui_server/js/CCCHelper.js
Endimion75: corrected

data/ui_server/js/CCCHelper.js
Endimion75: My intentions here is to create and empty array with the exact dimensions I require.  That way if I make a mistake down the road when storing the data in the right dimension, I will get an error.  If I use dynamic allocation I am afraid it would be too easy for me to make a mistake.

data/ui_server/js/CCCHelper.js
Endimion75: I am not comfortable doing refactoring in an untyped language where it is very easy to get things wrong after the refactoring and because I don't have unit testing, it will be a pain, to make sure the refactoring was successful.

data/ui_server/js/CCCHelper.js
Endimion75: I chose to use the string pattern testing, because if I use colours other than basic ones, I only need to make one test, rather than three (one for each channel).
In my opinion it gives you more flexibility to change the colours and easily read the code.

data/ui_server/js/CCCHelper.js
Endimion75: my version of the code is different, so I guess I already fixed it or I pushed an old version :/

data/ui_server/js/CCCHelper.js
Endimion75: You are correct here ... I have no idea why I did this ...

Can you remind me what is the difference between blob and image as the response from the API is concerned?

data/ui_server/js/cccCanvasColonyMarker.js
Endimion75: Fixed
although for people like me (suck at math) it is more intuitive to read as it was.

data/ui_server/js/cccCanvasColonyMarker.js
Endimion75: If the condition is not meat, there is no need for correction

data/ui_server/js/cccCanvasColonyMarker.js
Endimion75: fixed

data/ui_server/js/cccCanvasColonyMarker.js
Endimion75: I am not familiar with bind and I had a ton of trouble getting the event listeners to work properly, so change it at your own risk ;)

data/ui_server/js/cccCanvasColonyMarker.js
Endimion75: this piece of code is no more

data/ui_server/js/cccCanvasColonyMarker.js
Endimion75: changed it

data/ui_server/js/CCCHelper.js
Endimion75: Alright, I changed it to use imageMin and ImageMax as the functions is part of a function that is used to render the image.

One important question though.  Why do the colony detection API has a blob max and min if the blob data is only true/false?

data/ui_server/js/cccCanvasColonyMarker.js
local-minimum: (just taking notes) fixed

data/ui_server/js/cccCanvasColonyMarker.js
local-minimum: Fixed

local-minimum:
I will close this PR now, because the merging will be made from #169

PR #142, opened 2017-08-09
CCC poly construction

local-minimum:
There's still an unstableness with the test data that may reflect instability in the solutions for the polynomial (that there can be local optima where the solution gets stuck if not enough and varied data is supplied). There are several validation-steps so it should be fairly risk-free to use but may cause test to fail at random.


scanomatic/data_processing/calibration.py
```diff
@@ -117,6 +119,8 @@ class CellCountCalibration(Enum):
     """:type : CellCountCalibration"""
     independent_data_source = 8
     """:type : CellCountCalibration"""
+    deployed_polynomial = 9
+    """:type : CellCountCalibration"""
```
local-minimum: Otherwise they `CellCountCalibration.deployed_polynomial` looks like an `int` to some linters (pycharm's in particular) while they in fact are enums.

scanomatic/data_processing/calibration.py
```diff
@@ -459,7 +469,8 @@ def add_image_to_ccc(identifier, image):
     image.save(Paths().ccc_image_pattern.format(identifier, im_identifier))
 
     ccc[CellCountCalibration.images].append(im_json)
-    _save_ccc_to_disk(ccc)
+    if not save_ccc_to_disk(ccc):
+        False
```
local-minimum: yes

scanomatic/data_processing/calibration.py
```diff
@@ -897,6 +908,49 @@ def load_data_file(file_path=None, label=''):
...
+
+
+    return {
+        CalibrationEntry.target_value:
+            target_value[inclusion_filter].tolist(),
+        CalibrationEntry.source_values:
+            np.array(source_values)[inclusion_filter].tolist(),
+        CalibrationEntry.source_value_counts:
+            np.array(source_value_counts)[inclusion_filter].tolist(),
+    }
```
local-minimum: The enums are part of python core from 3.4, I think and this is simply a backwards compatible implementation of that. But yes, namedtuples and enums are kind of similar things.

scanomatic/data_processing/calibration.py
```diff
@@ -1179,5 +1238,45 @@ def add_external_data_to_ccc(identifier, data_file, report):
     ccc[CellCountCalibration.independent_data] = (
         meta_data.get_column_index_from_all_plates(-1))
 
-    _save_ccc_to_disk(ccc)
-    return True
+    return save_ccc_to_disk(ccc)
+
+
+@_validate_ccc_edit_request
```
local-minimum: true, but there's no such system in SoM and this is a token for editing a specific ccc-file and I think I would want it to be like that even with other auth-systems... I'm not sure is what I'm saying.

scanomatic/data_processing/calibration.py
local-minimum: yes in fact it might be true and that would resolve the other issue with the dots in the keys in the json, but changing this systematically (for all the enums basically) would rewrite most lines in the module... maybe separate PR?

local-minimum:
I've updated two of the things you commented on, but the the fixing the return -> throw exception and fixing the namedtuple suggestion, while they sound reasonable, I think they would cause changes in so many places that it is better to do it in separate PR

gdetrez:
> I've updated two of the things you commented on, but  the the fixing the return -> throw exception and fixing the  namedtuple suggestion, while they sound reasonable, I think they would  cause changes in so many places that it is better to do it in separate  PR

Yes, I don't think you should refactor all of them, that would produce a huge PR and add little value at this point. But maybe just not introduce new ones... Anyway, the most important thing is to fix the tests

gdetrez:
I think the code changes look better now.
I'm not sure about the disabled test. If the test or the test data are the problem, isn't it possible to fix them? Unless there is some pressure to land the feature quickly, I'd rather have a working test...

local-minimum:
I think the time-budget that we have for implementing the entire CCC feature is getting strained if we spend more time on making a full data to test with at the moment since it is a couple of hours of copy paste at best. If we can land this, and perhaps open issue to revisit the test, then we can soon use the interface to gather real data to use in the test and hopefully solve it that way. Hopefully. But I agree that in a nice world we'd fix the test.

I propose I make a task for it and then if the code was good enough, we merge the pull request.

Issue #140, opened 2017-07-07
Update API endpoints to use http abort codes instead of success=False/True


skymandr:
Some endpoints have already been updated to use `jsonify` for successful calls and `json_abort` with appropriate error code for failed, but most remain. There is also the issue, that all endpoints still use both `success=False/True` and return a status code.

DoD: API calls no longer use the `success=` paradigm. Failed calls return appropriate status codes.

skymandr:
There is also the issue, at least in the `calibration_api.py`, that some failures assume authentication failure, but are not sure about it. These should be made sure, so that the status code is meaningful.

local-minimum:
I don't think this has to be an open issue anymore. We will work towards this but we will not do a focused replacement.

PR #143, opened 2017-08-17
T688 issue 140 api should use status codes

skymandr:
This is a partial pull request for #140, since this will be a rather large one. The affected files will likely not change much, but I will not merge until all are done. Please have a look at, in particular, whether the status codes used are appropriate in all cases, and if the returned messages are informative.

### DIFF
--- a/scanomatic/ui_server/analysis_api.py
+++ b/scanomatic/ui_server/analysis_api.py
@@ -9,7 +9,7 @@ from scanomatic.io.paths import Paths
 from scanomatic.models.factories.analysis_factories import AnalysisModelFactory
 from scanomatic.models.analysis_model import DefaultPinningFormats
 from scanomatic.image_analysis.grid_array import GridArray
-from .general import get_image_data_as_array
+from .general import get_image_data_as_array, json_abort
 
 
 def add_routes(app):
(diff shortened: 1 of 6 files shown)


local-minimum:
@Endimion75 I'm not sure exactly what is preferred according to you here and I get that whichever way we go is going to be annoying to say the least. So I'd like your opinion. As the API worked before upon doing a request, the JS-side needed to check if `success=True` on the 200-code/the non-error side and if not handle the fail there as well as having possibly quite similar error handling code for dealing with e.g. 40X-status messages from the server. The work we started was to make all fails return something else than 20X, probably 40X messages and hence simplify the handling of non-successes. There's possibly a whole bunch of problems with switching over, so we could do it in different ways:

1) Revert all changes done so far and await that SoM is only developed internally here to make changes
2) Make partial changes per API-section, which will mean that we have two non-compatible paradigms at the same time
3) Make complete changes slowly recuperate the functionality of the UI by dealing with updating js-code.

skymandr:
There might be some stuff in here that is worth keeping, but I doubt it. Suggest we delete this PR.

local-minimum:
It has grown stale so I agree with deleting and improving code as we go along

PR #144, opened 2017-08-22
T724 add ccc finalize endpoint

skymandr:
Adding an endpoint for allowing finalizing CCC editing. Please check in particular that the test coverage covers the acceptance criteria:

Given that the user provides the correct access-token
Given that the CCC is under construction
Given that there is a polynomial constructed
When the user requests to finalize the CCC
The CCC is made available to the system and can no longer be edited.

Remaining:
Endpoint test


skymandr:
If I understand correctly deffcaa should do all that is needed to make the user data persistent and accessible to the user. In particular, `docker-compose.override.yml` does not need any special attendance, correct? If I've missed something, or if there are hidden problems, then please tell me.

scanomatic/data_processing/calibration.py
```diff
@@ -325,12 +382,19 @@ def add_ccc(ccc):
 def activate_ccc(identifier):
 
     ccc = __CCC[identifier]
+    try:
+        if not has_valid_polynomial(identifier):
```
gdetrez: I'm also wondering if that's the right place for that. If I understand correctly, the `ccc` thing that is activated here is retrieved from disk. Wouldn't it be more appropriate to check the format when it is written?

scanomatic/data_processing/test/test_calibration.py
```diff
@@ -141,14 +140,20 @@ def data_store_bad_ccc(edit_bad_slope_ccc):
     return calibration._collect_all_included_data(edit_bad_slope_ccc)
 
 
+@pytest.fixture(scope='function')
```
gdetrez: This is not true: the fixture is added to a global dictionary, so it will be added to the whole test session.

scanomatic/data_processing/calibration.py
```diff
@@ -206,11 +217,16 @@ def wrapped(identifier, *args, **kwargs):
 
                 else:
 
-                    _logger.error("Can only modify the CCC {0} because it is not under construction".format(identifier))
+                    _logger.error(
+                        "Can not modify the CCC " +
```
skymandr: I'm ambivalent on the issue of breaking strings. I guess we should formulate a standard for the project.

Interesting that you want line break before binary operator! I prefer it that way as well, as does Donald Knuth and pep allows it, but our linter complains if I do, I believe.

scanomatic/data_processing/calibration.py
```diff
@@ -220,7 +236,48 @@ def wrapped(identifier, *args, **kwargs):
...
+    This is further to say, that a returned `None` is not really a
+    guarantee that the problem is with the access token, though that is
+    likely.  This is confusing, I know..."""
+    return True
+
+
+def is_valid_polynomial(polynomial):
+    """Checks that the polynomial is of the correct format, but _not_
+    that it is reasonable, which is taken care of by the function
+    `validate_polynomial()`"""
```
skymandr: good point. will try to think of something.

scanomatic/data_processing/calibration.py
```diff
@@ -220,7 +236,48 @@ def wrapped(identifier, *args, **kwargs):
...
+                len(polynomial['coefficients']) == polynomial['power'] + 1)):
+            _logger.error(
+                "Validation of polynomial {} failed".format(polynomial))
+            raise ActivationError("Invalid polynomial: {}".format(polynomial))
+    except (KeyError, TypeError) as err:
+        _logger.error(
+            "Validation of polynomial failed with {}".format(err.message))
+        raise ActivationError(err.message)
+
+    return True
```
skymandr: I agree, it was a clumsy way of trying to use exceptions in a framework where booleans are the norm. I'll try to find a better way of doing it.

scanomatic/data_processing/calibration.py
skymandr: 1) Passing the identifier is the norm in the file. Don't think I want to change that without good reason.

2) The acceptance tests for the case said to check before activation, but it might be the case that it should be checked before writing as well. @local-minimum?

scanomatic/data_processing/calibration.py
skymandr: I guess this also needs input from @local-minimum

gdetrez: The linter can probably be configured to our preference

scanomatic/data_processing/calibration.py
gdetrez: I don't think you should try to follow (implicit) norms if there is a better way to do it. (Just to be clear, IMO the better way here is the exceptions)

scanomatic/data_processing/calibration.py
gdetrez: > Passing the identifier is the norm in the file. Don't think I want to change that without good reason.
Like above, I don't think implicit norms are relevant. It would make the code simpler and avoid the need to access a global variable in the validation.

docker-compose.yml
```diff
@@ -2,4 +2,5 @@ scanomatic:
     build: .
     ports:
         - 5000:5000
-
+    volumes:
+        - $HOME/.scan-o-matic:/root/.scan-o-matic
```
gdetrez: I don't think it's a good idea to have the docker image leak in the user directory. Unless there's a very good reason not to, I suggest using named volumes

scanomatic/data_processing/test/test_calibration.py
skymandr: Suggestion?

scanomatic/data_processing/calibration.py
local-minimum: I'm pretty much with @gdetrez on the keep strings intact is more important than 80 char width, but then again I don't respect the 80 char width at all.

gdetrez:
Easiest is to remove the ccc from the `_CCC` global dictionary after the
tests. Best is to refactor the code to get rid of the `_CCC`.

scanomatic/data_processing/calibration.py
skymandr: I removed the booleans in favour of using only exceptions. See new commit 4b1b852.

scanomatic/data_processing/calibration.py
skymandr: Ok, will fix!

scanomatic/data_processing/calibration.py
local-minimum: I'm getting used to this view while I still default to _keep with surrounding code pattern_ and while it is a bit annoying, it seems more reasonable to go for good solution even if it would look a bit different than everything else.

local-minimum:
Quick comments to some of the questions

scanomatic/data_processing/calibration.py
skymandr: Updated name, hopefully for the better. See new commit 4b1b852.

scanomatic/data_processing/calibration.py
skymandr: Changing this would mean also updating the wrapper (see `199 def _validate_ccc_edit_request(f):`) and consequently the functions using it. It doesn't seem difficult, but would require some work to make sure that it doesn't break anything in the calibration api. At a glance I don't see any reasons why this refactoring couldn't be done, but maybe it's outside the scope of this pull request (thought perhaps not of this task..?)?

scanomatic/data_processing/calibration.py
skymandr: Fixed in a07d901. What standard should we adopt for binary operators? I vote Knuth!

docker-compose.yml
skymandr: Ok, will have a look at that. Might need some assistance.

skymandr:
> Easiest is to remove the ccc from the `_CCC` global dictionary after the tests. Best is to refactor the code to get rid of the `_CCC`.

I agree that refactoring would be nice, and perhaps should be done within this work, but I have a feeling it would be best done as a separate pull request.

scanomatic/data_processing/calibration.py
gdetrez: I'm not sure I follow, it doesn't seems that `has_valid_polynomial` is decorated with `_validate_ccc_edit_request` here.

gdetrez:
>> Easiest is to remove the ccc from the _CCC global dictionary after the tests. Best is to refactor the code to get rid of the _CCC.

> I agree that refactoring would be nice, and perhaps should be done within this work, but I have a feeling it would be best done as a separate pull request.

The first option is not a big change and will at least prevent the fixture from leaking in other tests.

scanomatic/data_processing/calibration.py
skymandr: Right, thought you were talking about `activate_ccc()` rather than `has_valid_polynomial()`. Followed your advice, see commit 9f36c74.

docker-compose.yml
skymandr: Using named volumes, see commit 33a39e0.

scanomatic/data_processing/calibration.py
```diff
@@ -246,29 +246,27 @@ def is_valid_token(identifier):
...
+                err.message))
         raise ActivationError(err.message)
 
-    return True
-
 
 def has_valid_polynomial(identifier):
     try:
-        return is_valid_polynomial(
+        return is_valid_polynomial_repr(
```
gdetrez: I think you can drop the return here.

scanomatic/data_processing/calibration.py
```diff
@@ -246,29 +246,27 @@ def is_valid_token(identifier):
     return True
 
 
-def is_valid_polynomial(polynomial):
-    """Checks that the polynomial is of the correct format, but _not_
-    that it is reasonable, which is taken care of by the function
-    `validate_polynomial()`"""
+def is_valid_polynomial_repr(polynomial):
```
gdetrez: `repr` has an other meaning in python. Maybe `format` or `structure` instead? And since it doesn't return booleans anymore, maybe use `validate` instead of `is_valid`.

scanomatic/data_processing/calibration.py
skymandr: Fixed in commit a2ad337

scanomatic/data_processing/calibration.py
skymandr: Fixed in commit a2ad337

skymandr:
> Easiest is to remove the ccc from the `_CCC` global dictionary after the tests. Best is to refactor the code to get rid of the `_CCC`.

Fixed in commit 7f5c71a

skymandr:
> I would like to have tests for the views (at least the one you've added). Flask has documentation on how to set it up.

Implemented in commit 3c31571.

PR #148, opened 2017-08-29
Removed 'local' in factory since it is no longer in model

local-minimum:
Caused error trying to copy models since serialization was looking
for attribute local on the model which was no longer there

Also made lines shorter

### DIFF
--- a/scanomatic/models/factories/settings_factories.py
+++ b/scanomatic/models/factories/settings_factories.py
@@ -65,7 +65,6 @@ class UIServerFactory(AbstractModelFactory):
     STORE_SECTION_SERIALIZERS = {
         "port": int,
         "host": str,
-        "local": bool,
         "master_key": str,
     }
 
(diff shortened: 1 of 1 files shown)


scanomatic/models/factories/settings_factories.py
```diff
@@ -162,8 +163,10 @@ class ApplicationSettingsFactory(AbstractModelFactory):
         "computer_human_name": str,
         "scanner_models":
             lambda enforce=None, serialize=None:
-            ([serialize[name] for name in sorted(serialize.keys())] if isinstance(serialize, dict) else None)
-            if serialize is not None else (enforce if not isinstance(enforce, tuple) else list(enforce)),
+            ([serialize[name] for name in sorted(serialize.keys())]
```
skymandr: So hanging indent is OK in this repo? I think it has its use, but is strongly discouraged for internal projects by the boss! =)

skymandr: I would suggest defining this lamda function outside this dictionary for readabilty. No functional changes required, but it is very hard to read with the kind of linebreaks etc. that are required to define it here.

skymandr:
LGTM

PR #154, opened 2017-09-05
merging from scanning repo

dindinoh:


Dockerfile
```diff
@@ -1,4 +1,10 @@
-FROM python:2.7
+FROM ubuntu:16.04
+RUN apt update && apt -y install usbutils software-properties-common python-pip
+RUN echo -e "\n" | add-apt-repository ppa:rolfbensch/sane-release
+RUN apt-get update
+RUN apt -y install libsane=1.0.27-xenial1 sane-utils=1.0.27-xenial1 libsane-common=1.0.27-xenial1
+RUN echo "usb 0x4b8 0x12c" >> /etc/sane.d/epson2.conf
```
skymandr: What do these lines actually accomplish? Maybe a comment would be good, in case we need to do it (whatever it is) for other scanners in the future?

Dockerfile
```diff
@@ -1,4 +1,10 @@
-FROM python:2.7
+FROM ubuntu:16.04
+RUN apt update && apt -y install usbutils software-properties-common python-pip
+RUN echo -e "\n" | add-apt-repository ppa:rolfbensch/sane-release
+RUN apt-get update
```
skymandr: Maybe a minor detail, but build times might be quicker if we only do `apt update` once. (Might also take longer, depending on whether we get the benefit of caching - think on it!)

skymandr: To be clear, this is not a requested change, but I request that you think about what is best for clarity, efficiency, maintainabilty etc.

Dockerfile
skymandr: ... this, however, is a request.

Issue #149, opened 2017-08-31
Volume as endpoint for data


dindinoh:
Make both scan-o-matics use static volume as mount in docker.
Make it so docker mounts a volume for images.
Remove UI setting for choosing storage point.

dindinoh:
changed description of issue according to @skymandr comment https://github.com/Scan-o-Matic/scanomatic/pull/155#pullrequestreview-60848459

skymandr:
Close?

PR #155, opened 2017-09-06
#149 add static storage for project root as /somprojects that in turn…

dindinoh:
… can be mounted as needed on host machine


docker-compose.yml
```diff
@@ -9,6 +9,7 @@ services:
     volumes:
       - scan-o-matic-home:/root/.scan-o-matic
       - /dev/bus/usb:/dev/bus/usb
+      - /somprojects:/somprojects
```
skymandr: A minor grievance, but `somprojects` looks like it is a misspelled `someprojects`. I find that a little confusing every time I see it, so far without considerable diminishing returns.

PR #158, opened 2017-09-11
Remove tags from scanning

local-minimum:


scanomatic/io/xml/writer.py
```diff
@@ -200,7 +192,7 @@ def write_header(self, meta_data, plates):
                 f.write('<project>')
 
                 f.write(self.XML_OPEN_CONT_CLOSE.format(
-                    ['version', 'ver'][use_short_tags], __version__))
+                    ['version', 'ver'][use_short_tags], get_version()))
```
local-minimum: are short tags not tags?

local-minimum: see also below

scanomatic/io/xml/writer.py
```diff
@@ -362,7 +354,8 @@ def write_header(self, meta_data, plates):
                 for compartment in self.COMPARTMENTS:
```
local-minimum: this is one deep for loop structure...

local-minimum: (if you change it, make sure to write a test for the function first.!)

scanomatic/ui_server/scan_api.py
```diff
@@ -89,30 +107,41 @@ def get_scan_instructions(project=None):
...
+            success=True,
+            is_endpoint=True,
+            models=sane.get_scanner_models())
 
     @app.route("/api/scan/sane/modes/<model>")
     def get_scanner_modes(model):
         modes = sane.get_scanning_modes(model)
         if not modes:
-            jsonify(success=False, is_endpoint=True, reason="Scanner model '{0}' unknown".format(model))
+            jsonify(
```
local-minimum: json abort?

scanomatic/ui_server/scan_api.py
```diff
@@ -89,30 +107,41 @@ def get_scan_instructions(project=None):
...
+                success=False,
+                is_endpoint=True,
+                reason="Scanner model '{0}' unknown".format(model))
 
         mode_to_text = {sane.SCAN_MODES.TPU: "Transparency",
                         sane.SCAN_MODES.TPU16: "16 bit Transparency",
                         sane.SCAN_MODES.COLOR: "Reflective Color"}
 
         jsonify(success=True, is_endpoint=True,
```
local-minimum: line breaks

scanomatic/io/xml/writer.py
```diff
@@ -25,14 +16,11 @@
 import scanomatic.io.logger as logger
 from scanomatic.io.paths import Paths
 from scanomatic.models.analysis_model import COMPARTMENTS, MEASURES
-#
-# CLASSES
-#
+from scanomatic import get_version
```
local-minimum: new function? tested?

local-minimum: newly used here, but not new, not really tested...

scanomatic/io/xml/writer.py
local-minimum: I'd pref not work on this module, because I think we could safely drop xml import/output

scanomatic/ui_server/scan_api.py
local-minimum: prefer in separate branch since it will have consequences on the UI too.

skymandr:
löoks good to me

skymandr:
test

PR #161, opened 2017-09-14
Add favicon

skymandr:
Adding a favicon


local-minimum:
looking good to me too

Issue #58, opened 2016-12-08
Enhanced feature extraction UI


local-minimum:
If there's a state/previous extraction, allow for rerunning extraction without overwriting current QC work if possible.

- [ ] Add checkbox 'Overwrite previous QC (default)' <> 'Keep previous QC if possible'

local-minimum:
The API accepts `try_keep_qc` boolean (`1` or `0`) flag as of `v1.4.132`. This will attempt to keep the QC filter if set to true.

PR #162, opened 2017-09-15
Add toggle to extraction UI for try to keep qc

local-minimum:


local-minimum:
This resolves #58

gdetrez:
Would be better with a test but maybe not worth blocking the PR.

skymandr:
Suggest squash and merge for now. Test will be easier to implement once we've landed some refactoring.

PR #166, opened 2017-09-25
refactored routes in ui_server.py for easier testing (and overview)

skymandr:
- Refactored `ui_server.py` to put all routing in functions for easier Flask testing.
- Moved routes pertaining to the different options in the UI to their own files.
- Moved API functionality for experiment and settings to separate files.

Manually tested by running the service and clicking around.


scanomatic/ui_server/settings_api.py
```diff
@@ -0,0 +1,50 @@
...
+            app_conf.save_current_settings()
+            return jsonify(
+                success=success,
+                reason=None if success else "Bad data for {0}".format(
+                    bad_data))
+        elif action:
+            return jsonify(success=False, reason="Not implemented")
+
+    @app.route("/fixtures", methods=['post'])
+    def _fixtures():
```
skymandr: This is weird to me, but it was like this when I found it. Is this dead code or is it intentional that post to `/fixtures/` should give same as get? If this is planned future functionality is there an issue for it, and if so, should it abort with "Not implemented" instead? (Ping @local-minimum.)

local-minimum: It looks like it should be removed. But at least test that you can make a fixture and select it from a drop-down after it is made once you've removed it.

Issue #145, opened 2017-08-23
test_no_growth_only_noise sometimes fails


skymandr:
The test `test_no_growth_only_noise` sometimes fails with e.g.
```
    def test_no_growth_only_noise():
    
        phenotyper_object = build_test_phenotyper()
        model = build_model(phenotyper_object, 4)
    
        data = {}
        filt = np.ones_like(model.times, dtype=bool)
        assign_linear_phase_phenotypes(data, model, filt)
    
        assert np.isfinite(data[CurvePhasePhenotypes.LinearModelSlope]), "Invalid slope"
        assert np.isfinite(data[CurvePhasePhenotypes.LinearModelIntercept]), "Invalid intercept"
>       assert np.allclose(data[CurvePhasePhenotypes.LinearModelIntercept], 17, atol=0.5), "Unexpected intercept"
E       AssertionError: Unexpected intercept
E       assert False
E        +  where False = <function allclose at 0x7fed9f338050>(16.237456555526897, 17, atol=0.5)
E        +    where <function allclose at 0x7fed9f338050> = np.allclose

scanomatic/data_processing/test/test_phase_phenotypes.py:53: AssertionError

```

This happens rarely and is due to the fact that the test uses randomly generated data. Possible solutions are to use a static random data set, or to run the test multiple times and require that it passes 9 out of 10 times or similar.

skymandr:
Fixed in PR #181 by setting random seed explicitly. Not completely watertight, but maybe good enough, or what do you say, @local-minimum?

skymandr:
Test still skipped, perhaps still random failing (we know, or strongly suspect, that setting random seed explicitly didn't quite stick in the pytest context).

PR #168, opened 2017-09-26
Keep qc toggle second try, now with a bit of more tests

local-minimum:


skymandr:
When looking at the diffs I mostly see lint-things and the new tests. Were there also some functional changes? If so, have they been lost in git-space..?

(The failing test is our old friend the randomly failing one of #145 - annoying, but nothing to do with this.)

skymandr:
It seems that when you merged with master your changes were reverted your feature branch since we reverted them on master earlier. @gdetrez has expressed interest in performing some git magic on your behalf tomorrow.

PR #176, opened 2017-10-17
Bug: No or bad signal in grayscale detection mishandled

local-minimum:
## Solution

It is hard to know exactly what caused the problem, but the contents of this merge ensures that if no spikes where found or if no deltas were finite, the function should fail gracefully and the failing should be handled correctly.

In attempting to recreate this, a side effect error of the original case with producing grid images when some of the expected data was missing occurred. This is also included in this PR.

## Tests

Added tests for the function `scanomatic.image_analysis.signal.get_signal_edges` that hopefully captures both the scenario of having useful signal and having at least two kinds of bad signal.

## Original Traceback
```
Process RpcJob-3:
Traceback (most recent call last):
  File "/usr/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/local/lib/python2.7/dist-packages/scanomatic/server/rpcjob.py", line 115, in run
    effector_iterator.next()
  File "/usr/local/lib/python2.7/dist-packages/scanomatic/server/compile_effector.py", line 116, in next
    self._analyse_image(self._compile_job.images[self._image_to_analyse])
  File "/usr/local/lib/python2.7/dist-packages/scanomatic/server/compile_effector.py", line 143, in _analyse_image
    image_model = first_pass.analyse(compile_image_model, self._fixture_settings, issues=issues)
  File "/usr/local/lib/python2.7/dist-packages/scanomatic/image_analysis/first_pass.py", line 47, in analyse
    _do_grayscale(compile_analysis_model, fixture_image)
  File "/usr/local/lib/python2.7/dist-packages/scanomatic/image_analysis/first_pass.py", line 87, in _do_grayscale
    image.analyse_grayscale()
  File "/usr/local/lib/python2.7/dist-packages/scanomatic/image_analysis/first_pass_image.py", line 402, in analyse_grayscale
    current_model.grayscale.values = image_grayscale.get_grayscale(self, current_model.grayscale)[1]
  File "/usr/local/lib/python2.7/dist-packages/scanomatic/image_analysis/image_grayscale.py", line 141, in get_grayscale
    return get_grayscale_image_analysis(im, grayscale_area_model.name, debug=debug)
  File "/usr/local/lib/python2.7/dist-packages/scanomatic/image_analysis/image_grayscale.py", line 158, in get_grayscale_image_analysis
    return detect_grayscale(im_p, gs)
  File "/usr/local/lib/python2.7/dist-packages/scanomatic/image_analysis/image_grayscale.py", line 241, in detect_grayscale
    grayscale['sections'])
  File "/usr/local/lib/python2.7/dist-packages/scanomatic/image_analysis/signal.py", line 80, in get_signal_edges
    right=np.nan)
  File "/usr/local/lib/python2.7/dist-packages/numpy/lib/function_base.py", line 1881, in interp
    return interp_func(x, xp, fp, left, right)
ValueError: array of sample points is empty

```

### DIFF
--- a/scanomatic/image_analysis/first_pass_image.py
+++ b/scanomatic/image_analysis/first_pass_image.py
@@ -1,4 +1,5 @@
 import time
+import traceback
 import itertools
 import numpy as np
 
@@ -402,7 +403,10 @@ class FixtureImage(object):
             current_model.grayscale.values = image_grayscale.get_grayscale(
                 self, current_model.grayscale)[1]
         except TypeError:
-            self._logger.error("Grayscale detection failed / caused TypeError")
+            self._logger.error(
+                "Grayscale detection failed due to: \n{0}".format(
+                    traceback.format_exc()))
+                    
             current_model.grayscale.values = None
 
         if current_model.grayscale.values is None:
--- a/scanomatic/image_analysis/first_pass_image.py
+++ b/scanomatic/image_analysis/first_pass_image.py
@@ -399,13 +399,16 @@ class FixtureImage(object):
             return False
 
         try:
-            current_model.grayscale.values = image_grayscale.get_grayscale(self, current_model.grayscale)[1]
+            current_model.grayscale.values = image_grayscale.get_grayscale(
+                self, current_model.grayscale)[1]
         except TypeError:
+            self._logger.error("Grayscale detection failed / caused TypeError")
             current_model.grayscale.values = None
 
         if current_model.grayscale.values is None:
-            self._logger.error("Grayscale detection failed")
             return False
+
+        return True
 
     def _set_area_relative(self, area, rotation=None, offset=(0, 0), issues={}):
 
--- a/scanomatic/image_analysis/image_grayscale.py
+++ b/scanomatic/image_analysis/image_grayscale.py
@@ -221,6 +221,8 @@ def detect_grayscale(im_trimmed, grayscale):
 
     if gs_l_diff < NEW_GS_ALG_L_DIFF_T:
 
+        _logger.info('Using new method')
+
         deltas, observed_spikes, observed_to_expected_map = signal.get_signal_data(
             para_signal_trimmed_im, up_spikes, grayscale,
             grayscale["length"] * NEW_GS_ALG_L_DIFF_SPIKE_T)
@@ -241,6 +243,10 @@ def detect_grayscale(im_trimmed, grayscale):
                                             grayscale['sections'])
 
             fin_edges = np.isfinite(edges)
+            if not fin_edges.any():
+                _logger.error("No finite edges found")
+                return None, None
+
             where_fin_edges = np.where(fin_edges)[0]
 
             if DEBUG_DETECTION:
@@ -251,13 +257,13 @@ def detect_grayscale(im_trimmed, grayscale):
             frequency = frequency[np.isfinite(frequency)].mean()
 
             if not np.isfinite(frequency):
-                _logger.critical("No frequency was detected, thus no grayscale")
+                _logger.error("No frequency was detected, thus no grayscale")
                 return None, None
 
             edges = signal.extrapolate_edges(edges, frequency, para_signal_trimmed_im.size)
 
             if edges.size != grayscale['sections'] + 1:
-                _logger.critical(
+                _logger.error(
                     "Number of edges doesn't correspond to the grayscale segments ({0}!={1})".format(
                         edges.size, grayscale['sections'] + 1))
                 return None, None
--- a/scanomatic/image_analysis/image_grayscale.py
+++ b/scanomatic/image_analysis/image_grayscale.py
@@ -221,7 +221,7 @@ def detect_grayscale(im_trimmed, grayscale):
 
     if gs_l_diff < NEW_GS_ALG_L_DIFF_T:
 
-        _logger.info('Using new method')
+        _logger.info('Using default grayscale detection method')
 
         deltas, observed_spikes, observed_to_expected_map = signal.get_signal_data(
             para_signal_trimmed_im, up_spikes, grayscale,
--- a/scanomatic/image_analysis/signal.py
+++ b/scanomatic/image_analysis/signal.py
@@ -57,12 +57,11 @@ def get_signal_data(strip_values, up_spikes, grayscale, delta_threshold):
     return np.array(deltas), observed_spikes, observed_to_expected_index_map
 
 
-def get_signal_edges(observed_to_expected_index_map, deltas, observed_spikes, number_of_segments):
-
-    _logger.info('observed_to_expected_index_map ' + list(observed_to_expected_index_map))
-    _logger.info('deltas ' + list(deltas))
-    _logger.info('observed_spikes ' + observed_spikes)
-    _logger.info('number_of_segments ' + number_of_segments)
+def get_signal_edges(
+        observed_to_expected_index_map,
+        deltas,
+        observed_spikes,
+        number_of_segments):
 
     edges = np.ones((number_of_segments + 1,)) * np.nan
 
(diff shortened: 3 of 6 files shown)


scanomatic/image_analysis/image_grayscale.py
```diff
@@ -221,6 +221,8 @@ def detect_grayscale(im_trimmed, grayscale):
 
     if gs_l_diff < NEW_GS_ALG_L_DIFF_T:
 
+        _logger.info('Using new method')
```
skymandr: This is a pretty useless log-message without any context. New method?

scanomatic/image_analysis/image_grayscale.py
```diff
@@ -241,6 +243,10 @@ def detect_grayscale(im_trimmed, grayscale):
                                             grayscale['sections'])
 
             fin_edges = np.isfinite(edges)
+            if not fin_edges.any():
+                _logger.error("No finite edges found")
```
skymandr: This one is probably useful, but it would be better if the log message came with more context, e.g. what it was trying to do.

scanomatic/image_analysis/image_grayscale.py
```diff
@@ -251,13 +257,13 @@ def detect_grayscale(im_trimmed, grayscale):
             frequency = frequency[np.isfinite(frequency)].mean()
 
             if not np.isfinite(frequency):
-                _logger.critical("No frequency was detected, thus no grayscale")
+                _logger.error("No frequency was detected, thus no grayscale")
```
skymandr: This one at least has a minimum of context, could still be better! :)

scanomatic/image_analysis/signal.py
```diff
@@ -73,11 +77,17 @@ def get_signal_edges(observed_to_expected_index_map, deltas, observed_spikes, nu
...
+        edge_ordinals = np.arange(edges.size, dtype=np.float) + 1
+        edges[nan_edges] = np.interp(
+            edge_ordinals[nan_edges],
+            edge_ordinals[fin_edges],
+            edges[fin_edges],
+            left=np.nan,
+            right=np.nan)
+
+    elif nan_edges.any():
+        _logger.warning("No finite edges")
```
skymandr: idem.

scanomatic/image_analysis/test/test_signal.py
```diff
@@ -0,0 +1,107 @@
...
+
+        observed_to_expected_index_map = np.array([
+            0, 0, 0, 0, 0, 0, 0, 1, 2, 2, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8, 8,
+            9, 9, 10, 10, 11, 11, 11, 12, 12, 13, 13, 14, 14, 14, 15, 15, 16,
+            16, 16, 17, 17, 17, 17, 18, 18, 18, 19, 19, 19, 20, 20, 20, 20,
+            20, 20, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 21, 22, 22, 22,
+            22, 22, 22, 22, 22, 22, 22, 23, 23, 23, 23, 23
+        ])
+
+        deltas = np.array([
```
skymandr: is there a reason for the weird formatting of this array? If so, I think it could do with a comment, if not then please initialise it in a simpler way!

scanomatic/util/analysis.py
```diff
@@ -83,25 +92,30 @@ def make_grid_im(im, grid, save_grid_name, marked_position=None):
             x = 0
             y = 1
 
-            for row in range(grid.shape[1]):
+            if grid is not None:
+                for row in range(grid.shape[1]):
```
skymandr: with all the ifs this section is very deep. perhaps at least save one level by making a generator instead of the double for?

scanomatic/image_analysis/image_grayscale.py
local-minimum: The context is that there was a logging of 'Using fallback method' further down and I figured this would make it matching. The `_logger` is inited so you get context from it (message prefixed by `'Analyze Grayscale'`.

scanomatic/image_analysis/image_grayscale.py
local-minimum: Not sure I follow, as above it has more context of where it appeared and then there's also context in the sequence of the messages in the log. Or are you saying it should output some of the data that caused this condition? That could be good, but what data would be useful and not just a huge numpy-array...?

scanomatic/image_analysis/image_grayscale.py
local-minimum: as above, or if your refer to something else, please specify it.

scanomatic/image_analysis/signal.py
local-minimum: idem

scanomatic/image_analysis/test/test_signal.py
local-minimum: you mean where the line-breaks? No or yes, make linter (80 char line) happy with minimum work effort because these tests were really annoying copy-paste replace to get the true data back to python format.

scanomatic/image_analysis/image_grayscale.py
skymandr: Ok, I'm happy with the context for all the logging messages if that prefix is there. All, that is to say, except this one. "new method" is too vague, not future proof (what is new might change) and doesn't really tell whoever reads the log anything useful. Either make it more specific or remove it.

scanomatic/image_analysis/test/test_signal.py
skymandr: The comment refers to the `deltas` arrray, filled with `nans`, which is unfortunately not shown in this small context window (69 and below).

scanomatic/util/analysis.py
local-minimum: The fors are not nested and I don't see how making them a single thing would help much, but I could break out the whole grid-part of the plotting to a function...

skymandr: Oh, I see. Well, I won't insist, but that would be nicer.

scanomatic/image_analysis/image_grayscale.py
local-minimum: fixed

scanomatic/image_analysis/test/test_signal.py
local-minimum: I figured, and that was the answer. Because original deltas had uneven number of elements per line as consequence of differently large numbers, then replace them with nans made lines uneven in length and thus the format. Will fix though.

scanomatic/util/analysis.py
skymandr: yep, that's nicer.

scanomatic/image_analysis/test/test_signal.py
skymandr: Better, but why not something like `np.ones(observed_to_expected_index_map.shape) * np.nan` or similar? Seems clearer and more maintainable to me.

scanomatic/image_analysis/image_grayscale.py
skymandr: Much better!

scanomatic/image_analysis/test/test_signal.py
local-minimum: fixed

scanomatic/image_analysis/first_pass_image.py
```diff
@@ -399,14 +399,17 @@ def analyse_grayscale(self):
             return False
 
         try:
-            current_model.grayscale.values = image_grayscale.get_grayscale(self, current_model.grayscale)[1]
+            current_model.grayscale.values = image_grayscale.get_grayscale(
+                self, current_model.grayscale)[1]
         except TypeError:
+            self._logger.error("Grayscale detection failed / caused TypeError")
```
skymandr: Any reason not to use `as err` and log the actual error rather than just `caused TypeError`?

local-minimum: Not really, was considering it but then I don't remember if I ever implemented an error-logging method and formatting errors to tracebacks manually is a bit annoying. Remember, SoM isn't using the logging-module. Could check though (because, yes it would have helped).

skymandr: OK. I'll wait with approving until you've had the time to investigate this then.

local-minimum: Figured out I way, I think.

PR #178, opened 2017-10-17
Runtime warning grid empty mean

local-minimum:
## Description

One of the grid functions could attempt mean of an empty slice. Fixed so it now treats this as a gridding failure. Unfortunately the function was used by a function that was used by a ... you get it. So I needed logic to all those functions and tests for those too.

## Tests

Many for all the different functions involved.
Some of them are a bit weird in what they test for lack of more direct signs of failing grid detection.
Also fixed `tox.ini`
Also some test using the hard image, could start to fail if we improve the gridding because the hard image isn't impossible, but then we just need a new harder/impossible fixture image.

## Original warning

```
/usr/local/lib/python2.7/dist-packages/scanomatic/image_analysis/grid.py:278: RuntimeWarning: Mean of empty slice.
  deltas < expected_delta * (1 + leeway))].mean()
```

### DIFF
--- a/scanomatic/image_analysis/grid.py
+++ b/scanomatic/image_analysis/grid.py
@@ -152,7 +152,7 @@ def get_grid_parameters(x_data, y_data, grid_shape, spacings=(54, 54)):
 
     data = (x_data, y_data)
     new_spacings = get_grid_spacings(x_data, y_data, *spacings)
-    if new_spacings is None:
+    if None in new_spacings:
         return None, None
     centers = get_centre_candidates(grid_shape, new_spacings)
     votes = get_votes(data, centers)
@@ -160,7 +160,8 @@ def get_grid_parameters(x_data, y_data, grid_shape, spacings=(54, 54)):
     sigma = np.max((spacings[0], new_spacings[1])) * 0.1 / np.sqrt(2) + 0.5
     heatmap = get_heatmap(data, votes, weights, sigma)
 
-    _center_dim2, _center_dim1 = np.unravel_index(heatmap.argmax(), heatmap.shape)
+    _center_dim2, _center_dim1 = np.unravel_index(
+        heatmap.argmax(), heatmap.shape)
 
     new_center = (_center_dim1, _center_dim2)
 
--- a/scanomatic/image_analysis/grid.py
+++ b/scanomatic/image_analysis/grid.py
@@ -446,7 +446,6 @@ def get_grid(
         x_data=None,
         y_data=None,
         expected_center=(100, 100),
-        run_dev=False,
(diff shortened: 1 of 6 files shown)


scanomatic/image_analysis/grid.py
```diff
@@ -273,9 +276,13 @@ def get_grid_spacings(x_data, y_data, expected_dx, expected_dy, leeway=0.1):
     # noinspection PyUnresolvedReferences
     def get_delta(data, expected_delta):
         deltas = np.abs(np.subtract.outer(data, data))
-        return deltas[np.logical_and(
+        filt = np.logical_and(
```
skymandr: Maybe `&` better than `np.logical_and` for readability?

skymandr: They don't do exactly the same things though, so I'm not quite sure what the consequences might be...

scanomatic/image_analysis/grid.py
```diff
@@ -273,9 +276,13 @@ def get_grid_spacings(x_data, y_data, expected_dx, expected_dy, leeway=0.1):
     # noinspection PyUnresolvedReferences
     def get_delta(data, expected_delta):
```
skymandr: should be `get_deltas` since it no longer returns a single delta?

scanomatic/image_analysis/grid.py
```diff
@@ -478,6 +496,9 @@ def get_grid(im, expected_spacing=(105, 105), grid_shape=(16, 24),
             center, spacings = get_grid_parameters(
```
skymandr: There is a lot of repetition in this `if`. Possible to move to outside?

scanomatic/image_analysis/grid.py
local-minimum: If they're not identical in their operation, I'd prefer not to change it.

scanomatic/image_analysis/grid.py
local-minimum: I think it only returns a single value (the mean or a `None`).

scanomatic/image_analysis/grid.py
skymandr: Fair enough. I don't know if and how they are different, but I think that there most likely are cases.

scanomatic/image_analysis/grid.py
skymandr: You're right.

scanomatic/image_analysis/grid.py
local-minimum: the `run_dev` is never used. So I remove it.

PR #181, opened 2017-10-20
CCC duplicate ini error message

skymandr:
By mistake I also happened to also include a fix for #145 in this one, sorry about that...


scanomatic/ui_server_data/js/ccc.js
```diff
@@ -307,13 +307,18 @@ window.executeCCC = function() {
 
     function initiateNewCcc() {
         var valid = true;
+        var validCharsRegexp = /^[a-z]([0-9a-z_\s])+$/i;
```
local-minimum: use `let` and `const` rather than var for all cases.
See summary:
https://stackoverflow.com/a/11444416
Fix in many places...

scanomatic/ui_server_data/js/ccc.js
```diff
@@ -192,11 +192,11 @@ window.executeCCC = function() {
             500);
     }
 
-    function checkLength(o, n, min, max) {
-        if (o.val().length > max || o.val().length < min) {
-            o.addClass("ui-state-error");
+    function checkLength(obj, min, max, field) {
```
local-minimum: consider just putting the functions in `cccFunctions` directly if they don't manipulate the closure-variables at the top of the `executeCCC` function., it is a bit of a rabbit hole though since you'd need to move `updateTips` too. So I get if you don't feel like it.

scanomatic/ui_server_data/js/ccc.js
```diff
@@ -208,16 +208,20 @@ window.executeCCC = function() {
         }
     }
 
-    function checkRegexp(o, regexp, n) {
-        if (!(regexp.test(o.val()))) {
-            o.addClass("ui-state-error");
-            updateTips(n);
+    cccFunctions.checkLength = checkLength;
+
+    function checkName(obj, regexp, message) {
```
local-minimum: same here.

scanomatic/ui_server_data/js/ccc.js
```diff
@@ -192,11 +192,11 @@ window.executeCCC = function() {
             500);
     }
 
-    function checkLength(o, n, min, max) {
-        if (o.val().length > max || o.val().length < min) {
-            o.addClass("ui-state-error");
+    function checkLength(obj, min, max, field) {
+        if (obj.val().length > max || obj.val().length < min) {
+            obj.addClass("ui-state-error");
             updateTips("Length of " +
```
local-minimum: Since you are working around here, this type of line-break isn't valid in airb'n'b, though we haven't said we should follow it, it makes sense to me to minimize variation in coding style between the project.

Do
```
updateTips(
  adsf +
  asd +
)
```

Though that type of string concatenation isn't valid, use template literals which are actually kind of supported nowadays: https://caniuse.com/#feat=template-literals

scanomatic/ui_server_data/js/specs/ccc.spec.js
```diff
@@ -100,5 +100,77 @@ describe('setGriddingSuccess', () => {
...
             data, scope
         );
-    })
+    });
+});
+
+describe('checkName', () => {
+
+  it('rejects names starting with a number', () => {
+    // returns false
```
local-minimum: use `spyOn` to make sure side-effects such as `setState` isn't called and can be monitored what they would have been called with and then just call your function in the `cccFunctions.yourFunction(...)`.

scanomatic/ui_server_data/js/ccc.js
skymandr: fixed, I think

scanomatic/ui_server_data/js/ccc.js
skymandr: Yea, I'll leave that for now.

scanomatic/ui_server_data/js/ccc.js
skymandr: Fixed.

scanomatic/ui_server_data/js/ccc.js
skymandr: Don't understand what this refers to.

scanomatic/ui_server_data/js/specs/ccc.spec.js
skymandr: Left for later.

scanomatic/ui_server_data/js/ccc.js
local-minimum: maybe one comment disappeared, this function and the above seemed to be not needing the variables kept in scope of the closure/only manipulate variables passed to the function and return a value. It could therefore be lifted out and placed directly in `cccFunctions`, but this would require moving `updateTips` which I haven't checked how easy that would be.

scanomatic/ui_server_data/js/ccc.js
```diff
@@ -195,13 +195,8 @@ window.executeCCC = function() {
...
             obj.addClass("ui-state-error");
-            updateTips("Length of " +
-                field +
-                " must be between " +
-                min +
-                " and " +
-                max +
-                ".");
+            updateTips(
+                "Length of ${field} must be between ${min} and ${max}.");
```
local-minimum: Should use "`"

PR #184, opened 2017-10-24
added buildstatus from Travis on Github

dindinoh:


local-minimum:
Nice, this means we could have tests-passing too I guess?

PR #185, opened 2017-10-24
Fix the fixture name input in Chrome

gdetrez:
Remove code that would take a fixture name `Foo bar`, turn it into `foo_bar.config`, then turn that into `Foo bar.config` to finally store it inside the `<input />` element to never use it again (and yes, that means `<input ...>Foo bar.config</input>` which is wrong).


gdetrez:
If you try a name like `foo/bar` it fails to create the fixture on the API (404) but the UI still shows the next step.

PR #191, opened 2017-11-01
Fix issue where "setting" a colony without background would lose the `scope`.

gdetrez:
"Fix" by refactoring the colony editor (for lack of a better name) to a react component.
Instead of storing the current colony data in the `scope` object, it uses local component state and doesn't have the reported problem.

Try to keep the UI mostly the same (DOM structure, styles) and reused most of the helper code (API and canvas manipulation).

Add test for the re-factored interaction (corresponding roughly to the removed jquery code) but not the re-used code (API and canvas drawing).

### DIFF
--- a/scanomatic/ui_server_data/js/ccc/components/ColonyEditor.js
+++ b/scanomatic/ui_server_data/js/ccc/components/ColonyEditor.js
@@ -32,13 +32,6 @@ export default class ColonyEditor extends React.Component {
     }
 
     render() {
-        const style = {
-            background: 'white',
-            display: 'inline-block',
-            paddingd: '3px',
-            textAlign: 'center',
-        };
-
         return (
             <div>
                 <div><span>Colony Image</span></div>
@@ -49,9 +42,7 @@ export default class ColonyEditor extends React.Component {
                 />
                 <div><br /></div>
                 <div><span>Colony MetaData</span></div>
-                <div style={style} >
-                    <ColonyFeatures data={this.state.data} />
-                </div>
+                <ColonyFeatures data={this.state.data} />
                 <button
                     className="btn btn-default btn-fix"
                     style={{ horizAlign: 'center' }}
--- a/scanomatic/ui_server_data/js/ccc/components/ColonyEditor.js
+++ b/scanomatic/ui_server_data/js/ccc/components/ColonyEditor.js
@@ -1,0 +1,77 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import ColonyFeatures from './ColonyFeatures';
+import ColonyImage from './ColonyImage';
+
+export default class ColonyEditor extends React.Component {
+    constructor(props) {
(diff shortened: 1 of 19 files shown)


scanomatic/ui_server_data/js/ccc/components/ColonyEditor.js
```diff
@@ -0,0 +1,77 @@
...
+        this.setState({ drawing: true });
+    }
+
+    handleUpdate(data) {
+        this.setState({ drawing: false });
+        this.props.onUpdate && this.props.onUpdate(data);
+    }
+
+    render() {
+        const style = {
```
local-minimum: since there are other style-thingies in the JSX consider giving it a more specific name for what it is styling.

local-minimum: It's a bit annoying that styling is done in the js but GU really want this working next couple of weeks so I think we need to cut all corners we are comfortable with cutting and I'm OK with this corner.

scanomatic/ui_server_data/js/ccc/containers/ColonyEditorContainer.js
```diff
@@ -0,0 +1,82 @@
...
+        this.state = {};
+        this.handleSet = this.handleSet.bind(this);
+        this.handleUpdate = this.handleUpdate.bind(this);
+    }
+
+    componentDidMount() {
+        const { scope, ccc, image, plate, row, col, accessToken } = this.props;
+        SetColonyDetection(
+            scope, ccc, image, plate, accessToken, row, col,
+            this.handleColonyDetectionSuccess.bind(this),
```
local-minimum: bind callbacks in constructor

scanomatic/ui_server_data/js/ccc/containers/ColonyEditorContainer.js
```diff
@@ -0,0 +1,82 @@
...
+            () => {},
+        );
+    }
+
+    componentWillReceiveProps(newProps) {
+        this.setState({ colonyData: null });
+        const { scope, ccc, image, plate, row, col, accessToken } = newProps;
+        SetColonyDetection(
+            scope, ccc, image, plate, accessToken, row, col,
+            this.handleColonyDetectionSuccess.bind(this),
```
local-minimum: idem

scanomatic/ui_server_data/js/ccc/components/ColonyEditor.js
gdetrez: I moved it to the component it styles.

scanomatic/ui_server_data/js/ccc/containers/ColonyEditorContainer.js
gdetrez: I think that's an optimization mostly relevant for callbacks used in `render()`

PR #192, opened 2017-11-03
Make plate regriddable on success

gdetrez:
Refactor gridding interaction to use React and fix the issue.
There is lots of details in individual commit messages (which I would have expected github to copy here but I guess it's arc which does that).

This makes all the interaction for a given plate implemented with react :tada: 

As before, I didn't change the style/markup and the interaction looks and feels mostly the same


scanomatic/ui_server_data/js/ccc/containers/PlateEditorContainer.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    handleColonyFinish() {
+        const { row, col } = this.state.selectedColony;
+        const [nCol, nRow] = this.props.pinFormat;
+        if (col < nCol - 1) {
+            this.setState({ selectedColony: { row, col: col + 1 } });
+        } else if (row < nRow - 1) {
+            this.setState({ selectedColony: { row: row + 1, col: 0 } });
+        }
```
local-minimum: What happens when the last colony on the plate is finished?

scanomatic/ui_server_data/js/ccc/containers/GriddingContainer.js
```diff
@@ -0,0 +1,98 @@
...
+import React from 'react';
+
+import Gridding from '../components/Gridding';
+import { SetGridding } from '../api';
+
+export default class GriddingContainer extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            rowOffset: 4,
```
local-minimum: I suppose this is because it makes it correct most often on our test image, but I kind of don't like it because it is a debug setting and it should probably be starting with `0`.

scanomatic/ui_server_data/js/ccc/components/PlateEditor.js
```diff
@@ -0,0 +1,70 @@
...
+import React from 'react';
+
+import GriddingContainer from '../containers/GriddingContainer';
+import ColonyEditorContainer from '../containers/ColonyEditorContainer';
+import PlateProgress from './PlateProgress';
+
+export default function PlateEditor(props) {
+    let title = "Step 2: Gridding";
+    if (props.step === 'colony') {
+        title = 'Colony Detection'
```
local-minimum: Maybe this should be `Step 3: ...`

scanomatic/ui_server_data/js/ccc/index.js
```diff
@@ -488,6 +488,7 @@ window.executeCCC = function() {
                     plateId={scope.Plate}
                     pinFormat={scope.PinFormat.map((i) => parseInt(i))}
                     accessToken={scope.AccessToken}
+                    onFinish={() => alert('Level completed!')}
```
local-minimum: :)

PR #199, opened 2017-11-08
Javascript coverage

gdetrez:
Generate coverage report for the javascript tests and send to coveralls.

It's a tad convoluted: you need to generate the javascript coverage in the lcov file format, convert it to coveralls' json format using a ruby script and then pass this file to the python coveralls client that merges it with the python coverage.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,4 +31,5 @@ data/config/rpc.admin
 /htmlcov
 /Scan_o_Matic.egg-info
 docker-compose.override.yml
-scanomatic/ui_server_data/js/ccc.js
+/scanomatic/ui_server_data/js/ccc.js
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -31,4 +31,5 @@ data/config/rpc.admin
 /htmlcov
 /Scan_o_Matic.egg-info
 docker-compose.override.yml
-scanomatic/ui_server_data/js/ccc.js
+/scanomatic/ui_server_data/js/ccc.js
```
local-minimum: I probably missed this some time, but why it this in the gitignore?

gdetrez: This is the compiled file.

PR #200, opened 2017-11-10
CCC API: Store cell count with the pixel distribution

gdetrez:


scanomatic/data_processing/calibration.py
```diff
@@ -160,6 +160,8 @@ class CCCMeasurement(Enum):
     """:type : CCCMeasurement"""
     source_value_counts = 2
     """:type : CCCMeasurement"""
+    cell_count = 3
```
local-minimum: Probably remove `included` if you are not using it now.

scanomatic/data_processing/calibration.py
```diff
@@ -795,69 +797,47 @@ def transform_plate_slice(identifier, image_identifier, plate_id):
...
-            CCCMeasurement.included] = included
-
-    else:
-        plate[CCCPlate.compressed_ccc_data][x][y][
-            CCCMeasurement.included] = included
-        plate[CCCPlate.compressed_ccc_data][x][y][
-            CCCMeasurement.source_value_counts] = counts
-        plate[CCCPlate.compressed_ccc_data][x][y][
-            CCCMeasurement.source_values] = values
+             CCCMeasurement.source_values: [],
```
local-minimum: It could perhaps have been better to change `plate` to be a dictionary that have `(x, y)` as key? I understand if you think it is out of scope, but since we now store `cell_count` alongside with the `source_values` and `source_value_counts` we really have no need to fill up with empty placeholders.

PR #202, opened 2017-11-13
Define the CCC API as a flask blueprint

gdetrez:
Remove one level of indentation and define the url prefix only once.

I also unified the url prefix for the CCC API: most endpoint used
`/api/calibration/...` but a few used `/api/data/calibration/...`. Now
they all use the former.

### DIFF
--- a/scanomatic/ui_server/calibration_api.py
+++ b/scanomatic/ui_server/calibration_api.py
@@ -3,7 +3,7 @@ from itertools import product
 from types import StringTypes
 import re
 
-from flask import jsonify, request, send_file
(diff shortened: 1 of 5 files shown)


scanomatic/ui_server/calibration_api.py
```diff
@@ -47,848 +47,839 @@ def get_int_tuple(data):
...
-            data_object = request.values
+    if not success:
+        return json_abort(
+            400,
+            reason="Possibly someone just beat you to that combination " +
+            "of species and reference!"
+        )
+
+    return jsonify(
+        success=True,
```
skymandr: `success=True` is deprecated?

PR #203, opened 2017-11-13
Clean-up CCC API responses

gdetrez:
Remove all the `success=True` and `is_endpoint=True` from the CCC API
responses and corresponding checks in javascript.


gdetrez:
> Looks good, much cleaner, but I notice that you've removed the error alerts. Does this mean hat the service will crash and/or not give the user feedback when something goes wrong, or is there an error callback somewhere that I'm not seeing?

If you look in `index.js` there is an error callback function for each success callback. Those alerts would never have happened anyway because `success` was always `true`.

PR #205, opened 2017-11-14
Refactor JS API functions related to uploading images

gdetrez:
- return ES6 promises
- remove `scope` from the passed arguments
- add unit tests
- factor common code

### DIFF
--- a/scanomatic/ui_server_data/js/ccc/api.js
+++ b/scanomatic/ui_server_data/js/ccc/api.js
@@ -18,6 +18,30 @@ var SetColonyDetectionPath = baseUrl + "/api/calibration/#0#/image/#1#/plate/#2#
 var SetColonyCompressionPath = baseUrl + "/api/calibration/#0#/image/#1#/plate/#2#/compress/colony/#3#/#4#";
 
 
+class API {
+    static get(url) {
+        return new Promise((resolve, reject) => $.ajax({
+            url,
+            type: 'GET',
+            success: resolve,
+            error: jqXHR => reject(JSON.parse(jqXHR.responseText)),
+        }));
+    }
+
+    static postFormData(url, formData) {
+        return new Promise((resolve, reject) => $.ajax({
+            url,
+            type: 'POST',
+            contentType: false,
+            enctype: 'multipart/form-data',
+            data: formData,
+            processData: false,
+            success: resolve,
+            error: jqXHR => reject(JSON.parse(jqXHR.responseText).reason),
+        }));
+    }
+}
+
 export function GetSliceImageURL(cccId, imageId, slice) {
     var path = GetSliceImagePath.replace("#0#", cccId).replace("#1#", imageId).replace("#2#", slice);
     return path;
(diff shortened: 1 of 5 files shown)


scanomatic/ui_server_data/js/ccc/api.js
```diff
@@ -18,6 +18,30 @@ var SetColonyDetectionPath = baseUrl + "/api/calibration/#0#/image/#1#/plate/#2#
...
+    static get(url) {
+        return new Promise((resolve, reject) => $.ajax({
+            url,
+            type: 'GET',
+            success: resolve,
+            error: jqXHR => reject(JSON.parse(jqXHR.responseText)),
+        }));
+    }
+
+    static postFormData(url, formData) {
```
local-minimum: The endpoints support posting json the normal way so this should only be used if there's a need for form data such as uploading images.

scanomatic/ui_server_data/js/ccc/api.js
```diff
@@ -101,7 +115,7 @@ export function InitiateCCC(species, reference, successCallback, errorCallback)
     });
 }
 
-export function SetCccImageData(scope, cccId, imageId, accessToken, dataArray, fixture, successCallback, errorCallback) {
+export function SetCccImageData(cccId, imageId, accessToken, dataArray, fixture) {
     var path = SetCccImageDataPath.replace("#0#", cccId).replace("#1#", imageId);
```
local-minimum: I would really like having these changed and placed with the functions if they're only used in one place each. The whole replace scheme too compared to `${cccId}` ...

scanomatic/ui_server_data/js/ccc/index.js
```diff
@@ -382,37 +384,39 @@ window.executeCCC = function() {
...
     }
 
     function setGrayScaleImageAnalysisSuccess(data, scope) {
         cccFunctions.setStep(1.7);
         //store target_values and source_values to QC graph ???
-        GetFixturePlates(scope.FixtureName, function(data) {
-            createFixturePlateSelection(data, scope);
-        });
+        GetFixturePlates(scope.FixtureName)
+            .then(data => createFixturePlateSelection(data, scope));
```
local-minimum: Would it not be nice with a `.catch((data) => alert(data.reason))` or similar for all those promises?

scanomatic/ui_server_data/js/ccc/api.js
gdetrez: I'm a bit split about this. On one hand, I do understand that it would make the API implementation simpler but it doesn't help the current task and sending form data is known to work.

scanomatic/ui_server_data/js/ccc/api.js
gdetrez: I have now moved the ones for which there are tests.

scanomatic/ui_server_data/js/ccc/index.js
gdetrez: Added a very simple catch in this case. I think the other ones are already handled.

PR #206, opened 2017-11-15
Refactor CCC UI to allow processing multiple images

gdetrez:
This is a first refactoring of the UI part that deals with uploading image for CCC.
It is functional and already allow the user to process multiple images but there are some things that remain to address:
- it only let you process the first plate, independently of how many plates there are in the fixture
- there might be a race condition between the plate transformation and colony detection, but only if the gridding works on the first try and if the user is very fast to click on the "Next-step" button

I'd also like to have it look less crappy if I have time :cowboy_hat_face: 

But since it's already a lot of changes, it might  be easier to review this part first :stuck_out_tongue_winking_eye:


scanomatic/ui_server_data/CCC.html
```diff
@@ -60,70 +60,8 @@
...
-                                    </div>
-                                </div>
-                            </div>
-                        </div>
-                    </div>
-                    <div id="react-root"></div>
-                </div>
-            </div>
         </div>
+        <div id="react-root"></div>
```
local-minimum: Not much left to remove now :)

scanomatic/ui_server_data/js/ccc/components/CCCEditor.js
```diff
@@ -0,0 +1,54 @@
...
+        );
+    } else {
+        const image = props.images[props.currentImage];
+        view = (
+            <PlateEditorContainer
+                pinFormat={props.pinFormat}
+                accessToken={props.accessToken}
+                cccId={props.cccId}
+                imageId={image.id}
+                plateId="1"
```
local-minimum: So if we need ability to have more than one plate per image, we need to deal with this hard-coded value?

scanomatic/ui_server_data/js/ccc/components/CCCEditor.js
```diff
@@ -0,0 +1,54 @@
...
+                onFinish={props.onFinishImage}
+            />
+        );
+    }
+
+    return (
+        <div>
+            <PlateList plates={props.images} />
```
local-minimum: This sounds more like an `ImageList` than a `PlateList`?

scanomatic/ui_server_data/js/ccc/components/ImageUpload.js
```diff
@@ -0,0 +1,28 @@
...
+export default class ImageUpload extends React.Component {
+    constructor(props) {
+        super(props);
+        this.handleFileChange = this.handleFileChange.bind(this);
+    }
+
+    handleFileChange(event) {
+        const { target: { files } } = event;
+        this.props.onImageChange(files[0]);
+    }
```
local-minimum: kind of breaks the container - component pattern doesn't it? but I see the point of having it here.

scanomatic/ui_server_data/js/ccc/components/PlateList.js
```diff
@@ -0,0 +1,22 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export default function PlateList(props) {
+    return (
+        <div>
+          <h4>Uploaded Images</h4>
+          <ul>
```
local-minimum: maybe just make it a list-group?
https://getbootstrap.com/docs/3.3/components/#list-group

scanomatic/ui_server_data/js/ccc/helpers.js
```diff
@@ -167,3 +170,26 @@ export function loadImage(url) {
...
+        ];
+        return API.SetCccImageData(ccc, imageId, token, imageData, fixture);
+    }).then(() => {
+        return API.SetCccImageSlice(ccc, imageId, token);
+    }).then(() => {
+        return API.SetGrayScaleImageAnalysis(ccc, imageId, token);
+    }).then(() => {
+        return imageId;
+    });
+}
```
local-minimum: :)

scanomatic/ui_server_data/js/ccc/index.js
```diff
@@ -3,80 +3,15 @@ import ReactDOM from 'react-dom';
...
 } from './api';
-import { createScope, getCurrentScope, setCurrentScope } from './scope';
-import PlateEditorContainer from './containers/PlateEditorContainer';
+
+import CCCEditorContainer from './containers/CCCEditorContainer';
 
 
 
 window.cccFunctions = {
-    setStep: (step) => {
```
local-minimum: Yay it is gone!

scanomatic/ui_server_data/js/ccc/components/CCCEditor.js
gdetrez: Yes.

scanomatic/ui_server_data/js/ccc/components/ImageUpload.js
gdetrez: I don't think so. Container are about state and side effects, this just massages data for the callback.

gdetrez:
I'm working on refactoring `PlateList` and supporting multiple plates, so I'm not sure it's worth fixing here...

PR #207, opened 2017-11-16
Create Polynomial Results UI

local-minimum:
Fixes some issues with the math in the backend.
Adds containers and components to visualize the results (the polynomial).
Best way to test it is to fix the detection of the two first colonies on the Hanna-plate and set values so the second one has a larger value than the first. This should return a valid polynomial.

A note about that the `<ScientificNotation />` tests are done on the html as text. I found it much easier to deal with since the built-in ways of testing directly against the jsx hid whatever wasn't matching the expectation due to the format.
The `<ScientificNotation />` only has nice formatting for values less than one million (for some reason), but this is way way out of the range of used values so I'd prefer not to spend more time on it.

`expect(wrapped.contains(<div />)).toBe(true);` would only output errors like `false != true` and nothing about what wasn't matching.


scanomatic/ui_server_data/js/ccc/index.js
```diff
@@ -465,15 +465,21 @@ window.executeCCC = function() {
...
-            <PlateEditorContainer
-                cccId={scope.cccId}
-                imageId={scope.CurrentImageId}
-                plateId={scope.Plate}
-                pinFormat={scope.PinFormat.map((i) => parseInt(i))}
-                accessToken={scope.AccessToken}
-                onFinish={() => alert('Level completed!')}
-            />,
+        ReactDOM.render((
+            <div>
```
gdetrez: This should probably also be a component...

scanomatic/ui_server_data/js/ccc/containers/ResultsContainer.js
```diff
@@ -0,0 +1,75 @@
...
+    handleConstructionResultsError(reason) {
+        this.setState({ error: reason });
+    }
+
+    handleClearError() {
+        this.setState({ error: null });
+    }
+
+    render() {
+        return <Polynomial
```
gdetrez: I don't really like the pattern here:
- the name `Polynomial` is not really descriptive
- the name `Results` is a bit vague as well (mabye `CalibrationResults` would be better?)
- there is yet an other component whose name is `PolyResults` :confused: 
- I think there should be a clearer connection between the name of the container and the name of the child component. Maybe you don't have to follow the pattern of calling your container `XxxContainer` but a way to mark them as related.

scanomatic/ui_server_data/js/ccc/components/PolyResults.js
```diff
@@ -0,0 +1,130 @@
...
+        </p>
+    );
+}
+
+PolynomialEquation.propTypes = {
+    coefficients: PropTypes.array.isRequired,
+};
+
+
+export function ScientificNotation({ value, precision }) {
```
gdetrez: this is a bit convoluted and I'm not sure it's correct.

I think a simpler way to get the exponent would be to `Math.round(Math.log10(value) / 3) * 3` and then the condition in the if statement becomes simply `exponent === 0`.

But it's late so I'm probably missing things. In any case, I suggest separating the math and the rendering in separate function to make the former easier to test.

scanomatic/ui_server_data/js/specs/containers/ResultsContainer.spec.js
```diff
@@ -0,0 +1,118 @@
...
+            power: -2,
+            polynomial: 'y = e ^ x',
+            resultsData: 'imaginenative',
+            error: 'nope',
+        };
+        wrapper.setState(state);
+        const poly = wrapper.find('Polynomial');
+        expect(poly.prop('power')).toEqual(state.power);
+        expect(poly.prop('polynomial')).toEqual(state.polynomial);
+        expect(poly.prop('data')).toEqual(state.resultsData);
```
gdetrez: Didn't you forget `error` here?

scanomatic/data_processing/calibration.py
```diff
@@ -66,14 +71,18 @@ def _ccc_edit_validator(identifier, **kwargs):
             else:
 
                 _logger.error(
-                    "Can not modify the CCC {0} because it is not under construction".format(
+                    "Can not modify {0} sincet not under construction".format(
```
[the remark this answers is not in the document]
local-minimum: fixed

scanomatic/data_processing/calibration.py
```diff
@@ -634,27 +624,34 @@ def coeffs():
...
 
     poly_vals = np.zeros((degree + 1))
     poly_vals[-2] = c1
     poly_vals[0] = cn
 
     _logger.info(
-        "Data produced polynomial {0} with 1 sigma per term (x^1, x^{2}) {1}".format(
-            poly_as_text(poly_vals), np.sqrt(np.diag(pcov)), degree)
+        "Produced polynomial {} with 1 sigma per term (x^1, x^{}) {}".format(
+            poly_as_text(poly_vals), degree, pcov)  # np.sqrt(np.diag(pcov)))
```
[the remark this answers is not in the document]
local-minimum: fixed

scanomatic/data_processing/test/test_calibration.py
```diff
@@ -674,3 +615,66 @@ def test_source_value_counts(self, measurement):
...
         assert measurement[calibration.CCCMeasurement.cell_count] == 1234
+
+
+class TestConstructPolynomial:
+
+    @pytest.mark.parametrize("data_store,power", (
+        (calibration.CalibrationData([], [], []), 5),
+        (calibration.CalibrationData([1], [[1]], [[1]]), 5),
+        (calibration.CalibrationData([1], [[1]], [[6]]), 5),
+        (calibration.CalibrationData([1], [[1, 4]], [[6, 2]]), 5),
```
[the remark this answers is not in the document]
local-minimum: fixed

scanomatic/data_processing/test/test_calibration.py
```diff
@@ -674,3 +615,66 @@ def test_source_value_counts(self, measurement):
...
+    ))
+    def test_too_little_data_raises(self, data_store, power):
+
+        with pytest.raises(calibration.CCCConstructionError):
+            calibration.calculate_polynomial(data_store, power).tolist()
+
+    def test_calibration_curve_fit_polynomial_function(self):
```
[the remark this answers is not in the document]
local-minimum: yes, thought of it but since each poly is tested for a bunch of scenarios it would be a bit awkward to parametrize, it could be two tests though, each of them parametrized?

scanomatic/ui_server_data/js/ccc/components/PolyConstructionButton.js
```diff
@@ -0,0 +1,16 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export default function PolyConstructionButton(props) {
```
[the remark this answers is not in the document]
local-minimum: Nope, removed (I was planning to make it more involved, to have the ability to set 3rd, 4th, or 5th polynomial (also explains why that power was all over the place)).

scanomatic/ui_server_data/js/ccc/components/PolyResults.js
```diff
@@ -0,0 +1,130 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+
+export default function PolyResults(
+    { onClearError, polynomial, data, error }
+) {
+    if (error) {
```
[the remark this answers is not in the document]
local-minimum: Fixed. I was feeling the same, but was exhausted by the volume of the PR.

scanomatic/data_processing/calibration.py
```diff
@@ -634,27 +624,34 @@ def coeffs():
...
+        raise CCCConstructionError("Invalid data (probably too little)")
 
     poly_vals = np.zeros((degree + 1))
     poly_vals[-2] = c1
     poly_vals[0] = cn
 
     _logger.info(
-        "Data produced polynomial {0} with 1 sigma per term (x^1, x^{2}) {1}".format(
-            poly_as_text(poly_vals), np.sqrt(np.diag(pcov)), degree)
+        "Produced polynomial {} with 1 sigma per term (x^1, x^{}) {}".format(
```
skymandr: What do you mean by `1 sigma per term`? You give pcov, which is the covariance matrix, I guess? Perhaps you meant to uncomment the commented part, but forgot?

scanomatic/ui_server_data/js/ccc/containers/ResultsContainer.js
local-minimum: I've renamed things to be
* `PolynomialConstruction`
* `PolynomialConstructionError`
* `PolynomialResultsInfo` (figure we could just as well put the future `PolynomialResultsPlot` in its own component)

This way it also matches the pattern of containers and components (made container `PolynomialConstructionContainer`)

scanomatic/ui_server_data/js/ccc/components/PolyConstructionButton.js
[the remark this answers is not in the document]
skymandr: There really is no reason why 5th order should always be the best, in particular when we go to other organisms than yeast, so I hope this feature is not forgotten about...

scanomatic/ui_server_data/js/specs/containers/ResultsContainer.spec.js
local-minimum: fixed

scanomatic/ui_server_data/js/ccc/index.js
local-minimum: I put it in the `CCCEditor`

scanomatic/ui_server_data/js/ccc/components/PolyResults.js
local-minimum: I redid removing the whole only even 3s exponents and put mathing in a function.
(To your previous question, in equations like this, no, it would be unusual to go by 3-exponents, but as a rule biologists like to not have scientific notation and favor the letters `m` `mu` `n` and such)

scanomatic/data_processing/calibration.py
local-minimum: No the output of leastsq is different from curvefit, so the commented bit was meant to be thrown out. However i didn't really understand the log comment so I just left it. Feel free to fix based on my pr

local-minimum:
I think I did all the comments so please have a look

scanomatic/data_processing/calibration.py
skymandr: In that do one of the following:
1) don't include `cov_x` in the log message and remove the mention of `1 sigma`,
2) include it, but write what it actually is (an estimate of the Jacobian of the solution, not very useful),
3) calculate the standard errors and include those (see example here, for instance: https://stackoverflow.com/questions/14581358/getting-standard-errors-on-fitted-parameters-using-the-optimize-leastsq-method-i )

gdetrez:
Github is hidding my comment! :imp: 
> @skymandr made a very good remark about the formatting: the e-notation (i.e. 1.23e+42) is relatively commonly understood in different environments so using it would make it easier to copy/paste the value in an other tool (like a python notebook or a spreadsheet).
> 
> It feels like, while a bit nicer to look at, this custom formatting might end-up burdening both users and maintainers.
> But you know the users best so I'm fine if you think it's best to keep it.

local-minimum:
The biologist in me (but I might be wrong) says that e-notation is the sort of thing that should never be displayed or presented to anyone. It is more or less and error.

But also, we can revert to that later if there's a cost of maintaining the solution. Right now it feels like wrong thing to spend time on given that we have a notation that is correct and is tested.

scanomatic/data_processing/test/test_phase_phenotypes.py
```diff
@@ -72,7 +72,7 @@ def test_no_growth_only_noise():
     assert np.isfinite(
         data[CurvePhasePhenotypes.LinearModelIntercept]), "Invalid intercept"
     assert np.allclose(
-        data[CurvePhasePhenotypes.LinearModelIntercept], 17, atol=0.5
+        data[CurvePhasePhenotypes.LinearModelIntercept], 17, atol=0.6
```
skymandr: See comment on the commit.

PR #208, opened 2017-11-17
adding repo name and travis changes to push successful master to dock…

dindinoh:
…er hub

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: python
 python:
     - "2.7"
+
+services:
+  - docker
+
 addons:
   apt:
     packages:
@@ -29,6 +33,12 @@ before_script:
 script:
     - tox
     - npm test
+    - docker-compose build
+
 after_success:
     - coveralls-lcov -v -n scanomatic/ui_server_data/coverage/report-lcov/lcov.info > js-coverage.json
     - coveralls --merge=js-coverage.json
+    - if [ "$TRAVIS_BRANCH" == "master" ]; then
+    docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD";
+    docker-compose push;
+    fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,4 @@ script:
 after_success:
     - coveralls-lcov -v -n scanomatic/ui_server_data/coverage/report-lcov/lcov.info > js-coverage.json
     - coveralls --merge=js-coverage.json
-    - if [ "$TRAVIS_BRANCH" == "master" ]; then
-    docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD";
-    docker-compose push;
-    fi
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -29,6 +33,9 @@ before_script:
 script:
     - tox
     - npm test
+    - docker-compose build
+
 after_success:
     - coveralls-lcov -v -n scanomatic/ui_server_data/coverage/report-lcov/lcov.info > js-coverage.json
     - coveralls --merge=js-coverage.json
+    - if [ "$TRAVIS_BRANCH" == "master" ]; then docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"; docker-compose push; fi
```
gdetrez: This doesn't do what one expects it to: `$TRAVIS_BRANCH` will be `master` for pull requests that target the master banch (https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables).

Maybe add `"$TRAVIS_PULL_REQUEST" == "false"`.

docker-compose.yml
```diff
@@ -2,6 +2,7 @@ version: '2'
 
 services:
   scanomatic:
+    image: phenomique/scanomatic-standalone:latest
```
gdetrez: Why not just `phenomique/scanomatic` ?

local-minimum: That was my suggestion, given that we think we'll split up the contents on scanomatic into different partial scanomatics...

gdetrez: I see. But why not still have `sanomatic` to be the complete version and have the patial ones as `scanomatic-ui`, `scanomatic-scanner`, etc (or however you want to split them)?

.travis.yml
dindinoh: True, that is confusing, at least to me. They have something called [Convenience Variables](https://docs.travis-ci.com/user/environment-variables/#Convenience-Variables) where one can find: 

>TRAVIS_PULL_REQUEST is set to the pull request number if the current job is a pull request build, or false if it’s not.

That seems more clear to me. How about using 

    if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then

instead?

gdetrez: I think, as suggested, that you need to add `"$TRAVIS_PULL_REQUEST" == "false"` to `"$TRAVIS_BRANCH" == "master"`. Otherwise you will push when someone pushes to other branches on the main repository (like you did here for instance :wink:)

docker-compose.yml
local-minimum: alright that might be equally good or better.

gdetrez:
I think this is still not correct.

Tips: if you look at the bottom of this page, you can see a "details" link next to the Travis checks. If you click on this links and scroll to the bottom of the log, you can see the output of your command, if any. 
In this case, there is two Travis check. I think the one ending with `/push` is triggered by pushing to a branch in this repo and the one ending with `/pr` is triggered by opening a pull request.

For once, it's actually useful to get both because you get the two cases that should *not* push to docker hub (one of them still does).

I still think the solution I suggested in the first comment would do the trick: check for **both** `"$TRAVIS_PULL_REQUEST" == "false"` and `"$TRAVIS_BRANCH" == "master"`.

PR #210, opened 2017-11-17
Refactor <PlateEditor /> as a bootstrap "panel"

gdetrez:
This is mostly UI changes but I also needed to lift the state from `<GriddingContainer />` into `<PlateEditorContainer />`.

It also fixes the (unlikely) race condition in `<PlateEditorContainer />` but still only allow processing 

Here's how it looks now (so that you don't have to run the system to see it :cowboy_hat_face:):
screenshot-2017-11-17 cell count calibration 3
screenshot-2017-11-17 cell count calibration 4
screenshot-2017-11-17 cell count calibration 6
screenshot-2017-11-17 cell count calibration 7


local-minimum:
This might not relate exactly to this PR but

* I would like the rings on the plate that show where the grid is to be around 50% larger radius or so.
* And that color for the background in the color-legend...

scanomatic/ui_server_data/js/ccc/api.js
```diff
@@ -155,9 +155,9 @@ export function SetGrayScaleTransform(cccId, imageId, plate, accessToken) {
     return API.postFormData(path, formData);
 }
 
-export function SetGridding(cccId, imageId, plate, pinningFormat, offSet, accessToken, successCallback, errorCallback) {
+export function SetGridding(cccId, imageId, plate, pinningFormat, offSet, accessToken) {
```
local-minimum: why not use `API.post...`

scanomatic/ui_server_data/js/ccc/components/PlateEditor.js
```diff
@@ -1,56 +1,122 @@
...
+    } else if (step === 'colony-detection') {
+        text = `${now}/${max}`;
+        className += ' label-primary';
+    } else if (step === 'done') {
+        text = 'Done!';
+        className += ' label-success';
+    }
+
+    return <span className={className}>{text}</span>
+}
```
local-minimum: No type-checking?

scanomatic/ui_server_data/js/ccc/components/PlateEditor.js
```diff
@@ -1,56 +1,122 @@
...
-                            ccc={props.cccId}
-                            image={props.imageId}
-                            plate={props.plateId}
-                            onFinish={props.onColonyFinish}
-                            row={props.selectedColony.row}
-                            col={props.selectedColony.col}
-                        />
-                    }
-                </div>
+        <div className="panel panel-default">
```
local-minimum: Not sure, but isn't this a bit long/complex return? Doesn't some stuff wish to be their own components?

scanomatic/ui_server_data/js/ccc/containers/PlateEditorContainer.js
```diff
@@ -44,25 +90,37 @@ export default class PlateEditorContainer extends React.Component {
     render() {
         return (
             <PlateEditor
-                step={this.state.step}
-                onGriddingFinish={this.handleGriddingFinish}
-                onColonyFinish={this.handleColonyFinish}
+                accessToken={this.props.accessToken}
```
local-minimum: could this be reduced with `{...this.props}` or is that a bit bad in that it is not explicit. Anyways, there's a lot of mapping between things that have identical names.

scanomatic/ui_server_data/js/ccc/components/PlateEditor.js
gdetrez: Any concrete suggestion? This component is already mostly just laying-out other components in different combination depending on the `step`. Maybe we could gain a few lines  by using something like `react-bootstrap` or creating our own components for react layout stuff. Not sure how to divide it otherwise...

scanomatic/ui_server_data/js/ccc/containers/PlateEditorContainer.js
gdetrez: Yeah. I thought about this but I couldn't decide if that's a good idea or not: On one hand, in particular for containers, it would make things much simpler if we passed `{...this.props}` and/or `{...this.state}`. On the other hand, as you pointed out before, it's possible to accidentally pass props that you didn't expect to pass that just happen to have the same name in the parent and the child component. And it gets even worst if you do that through several components...
A possible compromise could be to do that only in container component, to avoid accidentally cascading props through the component hierarchy.

I'm thinking it might also be a  smells, indicating that we (I) should refactor the state a bit. (For instance, `cccId`, `accessToken` and `pinFormat` could be gathered into a `cccMetadata` object and `imageId`, `imageName` and `plateId` could also be passed as one object, as they are in `<CCCEditor />` in #211.)
But since that would affect both PRs, I would rather not do it know :cold_sweat:

scanomatic/ui_server_data/js/ccc/api.js
gdetrez: 'cause you haven't merged it yet :-P

scanomatic/ui_server_data/js/ccc/components/PlateEditor.js
gdetrez: :ballot_box_with_check:

gdetrez:
> I would like the rings on the plate that show where the grid is to be around 50% larger radius or so.

:arrow_right: #213 

> And that color for the background in the color-legend...

:arrow_right: #212

scanomatic/ui_server_data/js/ccc/api.js
local-minimum: :+1:

scanomatic/ui_server_data/js/ccc/containers/PlateEditorContainer.js
local-minimum: It sound like the best plan. Make nested structure, but make it in different PR

scanomatic/ui_server_data/js/ccc/components/PlateEditor.js
local-minimum: You are right, no clear answer. Just making another component tthat is the combination of  `Gridding` and `ColonyEditorContainer` will not make things much better. Some gains could be obtained if the state and the props weren't that flat structures. For example if `rowOffset` and `colOffset` were

```
offset: {
   rows: 3,
   cols: 0,
}
```

But that is better dealt with seperately.

PR #211, opened 2017-11-17
Multi plate CCC editor

gdetrez:
This removes the limitation where only the first plate per image could be processed.
The `<CCCEditor />` now queries the API for the number of plates in the current fixture.

It now creates one `<PlateEditorContainer />` per plate for each image. This won't make too much sense  with the plate editor implementation currently in master but was developed on top of #210 and uses the `collapse` props introduced there to collapse all panels but the current one.

Including #210, it should look like this, going through plates from top to bottom:
screenshot-2017-11-17 cell count calibration 9


scanomatic/ui_server_data/js/specs/containers/CCCEditorContainer.spec.js
```diff
@@ -3,7 +3,22 @@ import { shallow } from 'enzyme';
...
+        }
+        return this;
+    }
+
+    static resolve(value) {
+        const fake = new FakePromise();
+        fake.value = value;
+        return fake;
+    }
+}
```
local-minimum: It looks like half a promise :) but maybe that's all that was needs (no `reject` and `catch`)?

gdetrez: I didn't need those.

gdetrez:
> `CCCEditorContainer.propTypes` has not been updated with `collapse` and `onFinish`.

Assuming you meant `PlateEditorContainer`. I added `collapse` but `onFinish` was already there.

PR #213, opened 2017-11-17
Increase radius of grid outlines by 50%

gdetrez:
Questions:
- Should the red dot be bigger as well?
- Doesn't it make the current-colony outline less visible?
- I don't find the images with a tighter pinning format. Will it still work with that?


local-minimum:
>   Should the red dot be bigger as well?

Honestly, I never truely liked the red dot since it hides the colony and since the grid image does show where you will start inspecting which is about the same info as the red dot has, I think it might be better to remove

>    Doesn't it make the current-colony outline less visible?

Yes unfortunately so. I was considering if line width of the current-colony could be increased and a splash of color too it too!

>    I don't find the images with a tighter pinning format. Will it still work with that?

Probably will look rather crazy, but I think we can fairly safely assume that the CCC will (almost) always be in 96-format

PR #214, opened 2017-11-20
Display upload progress when adding image to CCC

gdetrez:
- Update <ImageUpload /> to show a progress bar
- Update <ImageUploadContainer /> to update <ImageUpload /> progress
- Update uploadImage() helper to set progress using a callback

### DIFF
--- a/scanomatic/ui_server_data/js/ccc/components/ImageUpload.js
+++ b/scanomatic/ui_server_data/js/ccc/components/ImageUpload.js
@@ -13,10 +13,27 @@ export default class ImageUpload extends React.Component {
     }
 
     render() {
+        const progressWidth = this.props.progress
+            ? (this.props.progress.now / this.props.progress.max * 100)
+            : 0;
         return (
             <div className="ImageUpload">
                 <h3>Process new image</h3>
-                <input type="file" onChange={this.handleFileChange} />
+                {this.props.progress ?
+                    <div>
+                        <div className="progress">
+                            <div
+                                className="progress-bar"
+                                style={{ width: `${progressWidth}%` }} />
+                        </div>
+                        {this.props.progress.text}
+                    </div>
+                :
+                    <input
+                        type="file"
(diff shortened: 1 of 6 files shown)


scanomatic/ui_server_data/js/ccc/components/ImageUpload.js
```diff
@@ -13,10 +13,27 @@ export default class ImageUpload extends React.Component {
...
 
     render() {
+        const progressWidth = this.props.progress
+            ? (this.props.progress.now / this.props.progress.max * 100)
+            : 0;
         return (
             <div className="ImageUpload">
                 <h3>Process new image</h3>
-                <input type="file" onChange={this.handleFileChange} />
+                {this.props.progress ?
```
local-minimum: I think I'd prefer to make the jsx-return easier to parse by

```js
let contents = null;
if (this.props.progress) {
    contents = ...
} else {
    contents = ...
}

return (
    ...
    {contents}
    ...
);
```
Though perhaps with nicer names. Possibly do an `const contents = ? ... : ...;` instead.

scanomatic/ui_server_data/js/specs/containers/ImageUploadContainer.spec.js
```diff
@@ -65,4 +75,44 @@ describe('<ImageUploadContainer />', () => {
...
+        const wrapper = mount(<ImageUploadContainer {...props} />);
+        wrapper.find('ImageUpload').prop('onImageChange')(image);
+        const setProgress = helpers.uploadImage.calls.argsFor(0)[4];
+        setProgress(3, 4, 'You are here');
+        wrapper.update();
+        expect(wrapper.children().prop('progress'))
+            .toEqual({ now: 3, max: 4, text: 'You are here' });
+    });
+
+    it('should clearr the progress on upload success', () => {
```
local-minimum: typo `clearr` (at least also next function).

PR #217, opened 2017-11-21
Set jsx indentations to 4 spaces in eslint

gdetrez:


skymandr:
oh no! decreased coverage!

PR #225, opened 2017-11-23
Ccc calibration update

local-minimum:
Bit restructure to allow better testing and return more information from the polynomial construction

### DIFF
--- a/scanomatic/data_processing/calibration.py
+++ b/scanomatic/data_processing/calibration.py
@@ -686,34 +686,18 @@ def construct_polynomial(identifier, power):
 
     validation = validate_polynomial(slope, p_value, stderr)
 
-    if validation is not CalibrationValidation.OK:
-        return {
-            'validation': validation.name,
-            'polynomial_coefficients': poly_coeffs,
-            'measured_sizes': data_store.target_value.tolist(),
-            'calculated_sizes': calculated_sizes,
-            'polynomial_as_text': poly_as_text(poly_coeffs),
-            'correlation': {
-                'slope': slope,
-                'intercept': intercept,
-                'p_value': p_value,
-                'stderr': stderr,
-            },
-        }
-    ccc[CellCountCalibration.polynomial] = get_polynomal_entry(
-        power, poly_coeffs,
-    )
+    if validation is CalibrationValidation.OK:
+        ccc[CellCountCalibration.polynomial] = get_polynomal_entry(
+            power, poly_coeffs,
+        )
 
-    if not save_ccc_to_disk(ccc):
-        return False
+        if not save_ccc_to_disk(ccc):
+            return False
 
     calculated_sizes = calculate_sizes(data_store, poly)
 
     return {
-        'ccc': identifier,
         'polynomial_coefficients': poly_coeffs,
-        'polynomial_as_text': poly_as_text(poly_coeffs),
-        'polynomial_power': power,
         'measured_sizes': data_store.target_value.tolist(),
         'calculated_sizes': calculated_sizes,
         'validation': validation.name,
--- a/scanomatic/data_processing/calibration.py
+++ b/scanomatic/data_processing/calibration.py
@@ -477,6 +477,7 @@ def transform_plate_slice(identifier, image_identifier, plate_id):
 
     if not grayscale_targets or not grayscale_values:
         _logger.error("The gray-scale values have not been setup")
+        return False
 
     transpose_polynomial = Image_Transpose(
         sourceValues=grayscale_values,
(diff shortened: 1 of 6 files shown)


scanomatic/data_processing/calibration.py
```diff
@@ -683,11 +680,25 @@ def construct_polynomial(identifier, power):
...
+    calculated_sizes = calculate_sizes(data_store, poly)
+    slope, intercept, _, p_value, stderr = linregress(
+        data_store.target_value, calculated_sizes)
+
+    validation = validate_polynomial(slope, p_value, stderr)
 
     if validation is not CalibrationValidation.OK:
         return {
-            'validation': validation
+            'validation': validation.name,
```
gdetrez: This is almost copied verbatim bellow (`ccc` is omitted here but that's probably not intended). Remove duplication.

scanomatic/data_processing/calibration.py
```diff
@@ -701,10 +712,17 @@ def construct_polynomial(identifier, power):
     return {
         'ccc': identifier,
```
gdetrez: This is an argument to the function, why return it?

scanomatic/data_processing/calibration.py
```diff
@@ -701,10 +712,17 @@ def construct_polynomial(identifier, power):
...
         'polynomial_coefficients': poly_coeffs,
+        'polynomial_as_text': poly_as_text(poly_coeffs),
```
gdetrez: Avoid returning derived data. Call `poly_as_text` where you actually use it. Or maybe write a `Polynomial` class with a `__str__` method?

scanomatic/data_processing/calibration.py
```diff
@@ -701,10 +712,17 @@ def construct_polynomial(identifier, power):
...
         'polynomial_power': power,
```
gdetrez: This is also an argument.

scanomatic/data_processing/test/test_calibration.py
```diff
@@ -174,6 +177,10 @@ def test_construct_bad_polynomial(self, edit_bad_slope_ccc):
             response['validation'] is not
             calibration.CalibrationValidation.OK)
 
+        assert 'correlation' in response
+        assert 'measured_sizes' in response
+        assert 'calculated_sizes' in response
```
gdetrez: Those are not very interesting tests. Can't you predict the actual values here?

scanomatic/data_processing/test/test_calibration.py
```diff
@@ -185,6 +192,9 @@ def test_construct_good_polynomial(self, edit_ccc):
         response = calibration.construct_polynomial(
             identifier, power, access_token=token)
 
+        assert 'correlation' in response
+        assert response['correlation']['slope'] == pytest.approx(1, abs=0.02)
```
gdetrez: What about the other parts of the correlation?

scanomatic/ui_server/calibration_api.py
```diff
@@ -764,11 +764,18 @@ def construct_calibration(ccc_identifier, power):
...
+        extra_info += ' correlation: {}'.format(response['correlation'])
+
     return json_abort(
         400,
         reason="Construction refused. " +
-        "Validation of polynomial says: {}".format(
-            response["validation"].name
+        "Validation of polynomial says: {}{}".format(
+            response["validation"],
+            extra_info
```
gdetrez: Test?

scanomatic/data_processing/calibration.py
```diff
@@ -477,6 +477,7 @@ def transform_plate_slice(identifier, image_identifier, plate_id):
 
     if not grayscale_targets or not grayscale_values:
         _logger.error("The gray-scale values have not been setup")
+        return False
```
gdetrez: Why not use an exception?

local-minimum: Well none really except the method already used that pattern, but I can rewrite it so all cases raise and handle that raising in the endpoint if you want.

scanomatic/data_processing/calibration.py
local-minimum: fixed

scanomatic/data_processing/calibration.py
local-minimum: nope there's no point (fixed)

scanomatic/data_processing/calibration.py
local-minimum: yes (though causes ui fix) (fixed)

scanomatic/data_processing/test/test_calibration.py
local-minimum: fixed

scanomatic/data_processing/test/test_calibration.py
local-minimum: fixed

scanomatic/ui_server/calibration_api.py
local-minimum: not easy to make a nice test so I made something at least.

local-minimum: added what seemed doable...

PR #227, opened 2017-11-23
Refactor CCC initialization

gdetrez:


scanomatic/ui_server_data/js/ccc/api.js
```diff
@@ -56,17 +53,9 @@ export function GetSliceImage(cccId, imageId, slice, successCallback, errorCallb
 }
 
 
-export function GetFixtures(callback) {
-    var path = GetFixtruesPath;
-
-    d3.json(path, function(error, json) {
```
local-minimum: are there any d3 left in the ccc?

scanomatic/ui_server_data/js/ccc/components/CCCInitialization.jsx
```diff
@@ -0,0 +1,77 @@
...
+                                className="fixtures form-control"
+                                onChange={props.onFixtureNameChange}
+                                value={props.fixtureName}
+                            >
+                                {props.fixtureNames.map(v => (
+                                    <option key={v} value={v}>{v}</option>
+                                ))}
+                            </select>
+                        </div>
+                        <div className="form-group">
```
local-minimum: I don't know if it is time to consider re-usability here, but the selection of pinning format is done in 2-3 places on the site and so this could be a general component.

scanomatic/ui_server_data/js/ccc/components/CCCInitialization.jsx
```diff
@@ -0,0 +1,77 @@
...
+                            <label>Reference</label>
+                            <input
+                                className="reference form-control"
+                                placeholder="reference"
+                                value={props.reference}
+                                onChange={props.onReferenceChange}
+                            />
+                        </div>
+                        <div className="form-group">
+                            <label>Fixture</label>
```
local-minimum: The selection of fixture is done in serveral places in SoM so this could be a generic component, don't know if this is the time and place to work on that though...

scanomatic/ui_server_data/js/ccc/containers/CCCInitializationContainer.jsx
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    componentDidMount() {
+        GetFixtures().then(
+            this.handleGetFixturesSuccess.bind(this),
+            this.setError.bind(this, 'Error getting fixtures'),
+        );
+        GetPinningFormats().then(
+            this.handleGetPinningFormatsSuccess.bind(this),
+            this.setError.bind(this, 'Error getting pinning formats'),
```
local-minimum: If both these errors can happen more or less at the same time. Maybe not a great problem but I guess they don't stack errors, just replace them?

scanomatic/ui_server_data/js/ccc/containers/CCCInitializationContainer.jsx
```diff
@@ -0,0 +1,115 @@
...
+        for (let i = 0; i < this.state.pinningFormats.length; i += 1) {
+            if (this.state.pinningFormats[i].name === name) {
+                return this.state.pinningFormats[i];
+            }
+        }
+        return null;
+    }
+
+    handleGetFixturesSuccess(fixtureNames) {
+        this.setState({ fixtureNames, fixtureName: fixtureNames[0] });
```
local-minimum: How does this work if you have a success but there are no fixtures yet on the system? Will the component rendering handle this correctly?
It should probably direct the user to make new fixtures.

scanomatic/ui_server_data/js/ccc/containers/CCCInitializationContainer.jsx
```diff
@@ -0,0 +1,115 @@
...
+        if (!this.state.reference) {
+            this.props.onError('Reference cannot be empty');
+            return;
+        }
+        if (!this.state.fixtureName) {
+            this.props.onError('Fixture name cannot be empty');
+            return;
+        }
+        if (!this.state.pinningFormat) {
+            this.props.onError('Pinning format cannot be empty');
```
local-minimum: as above for these, will it handle nicely if all of them are wrong?

scanomatic/ui_server_data/js/specs/components/CCCInitialization.spec.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+    it('should call onReferenceChange when the reference changes', () => {
+        const wrapper = shallow(<CCCInitialization {...props} />);
+        const input = wrapper.find('input.reference');
+        const event = { target: { value: 'XXX' } };
+        input.simulate('change', event);
+        expect(onReferenceChange).toHaveBeenCalledWith(event);
+    });
+
+    it('should render a <select /> for the fixtures', () => {
```
local-minimum: about my comment above on handling system with no fixtures yet made, there needs to be test for that too ofc.

scanomatic/ui_server_data/js/ccc/api.js
gdetrez: Yes, but I think only the linear scale thingy is left.

scanomatic/ui_server_data/js/ccc/components/CCCInitialization.jsx
gdetrez: Well, you're the PO :stuck_out_tongue_winking_eye:

scanomatic/ui_server_data/js/ccc/containers/CCCInitializationContainer.jsx
gdetrez: Yes

scanomatic/ui_server_data/js/ccc/containers/CCCInitializationContainer.jsx
gdetrez: Hum, it'll probably crash. I can change it to show an error message.

scanomatic/ui_server_data/js/ccc/containers/CCCInitializationContainer.jsx
gdetrez: It depends how you define "nicely" :nerd_face: 
It will not cause a javascript error but only the first error message will be shown.

PR #228, opened 2017-11-24
Add a button to finalize CCC

gdetrez:
The finalize button is placed next to the construct button and enabled only if there's a polynomial.
Add a simple component that replaces the CCCEditor when the CCC is finalized.

### DIFF
--- a/scanomatic/ui_server_data/js/specs/components/FinalizedCCC.spec.jsx
+++ b/scanomatic/ui_server_data/js/specs/components/FinalizedCCC.spec.jsx
@@ -1,0 +1,29 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+
+import './enzyme-setup';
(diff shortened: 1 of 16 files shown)


scanomatic/ui_server_data/js/specs/components/FinalizedCCC.spec.jsx
```diff
@@ -0,0 +1,29 @@
...
+    const props = { cccMetadata };
+
+    it('should say "Well Done!"', () => {
+        const wrapper = shallow(<FinalizedCCC {...props} />);
+        expect(wrapper.text()).toContain('Well Done!');
+    });
+
+    it('should show the calibration as string', () => {
+        const wrapper = shallow(<FinalizedCCC {...props} />);
+        expect(wrapper.text()).toContain('S. Kombuchae, Professor X');
```
local-minimum: :)

PR #229, opened 2017-11-24
Add a <select /> to choose the degree of the polynomial

gdetrez:
- Use bootstrap inline form to arrange the select and the button
- Add a "degree of polynomial" label
- pass the selected degree to the api
- available values are 1 to 5 but that's easy to change

### DIFF
--- a/scanomatic/ui_server_data/js/ccc/components/PolynomialConstruction.js
+++ b/scanomatic/ui_server_data/js/ccc/components/PolynomialConstruction.js
@@ -25,7 +25,7 @@ export default function PolynomialConstruction(props) {
         );
     }
 
-    const degrees = ['1', '2', '3', '4', '5'];
+    const degrees = ['2', '3', '4', '5'];
     return (
         <div>
             <h3>Cell Count Calibration Polynomial</h3>
--- a/scanomatic/ui_server_data/js/ccc/components/PolynomialConstruction.js
+++ b/scanomatic/ui_server_data/js/ccc/components/PolynomialConstruction.js
@@ -60,7 +60,6 @@ PolynomialConstruction.propTypes = {
     onClearError: PropTypes.func.isRequired,
     onDegreeOfPolynomialChange: PropTypes.func.isRequired,
     polynomial: PropTypes.shape({
-        power: PropTypes.number.isRequired,
         coefficients: PropTypes.array.isRequired,
         colonies: PropTypes.number.isRequired,
     }),
--- a/scanomatic/ui_server_data/js/ccc/components/PolynomialConstruction.js
+++ b/scanomatic/ui_server_data/js/ccc/components/PolynomialConstruction.js
@@ -42,7 +42,7 @@ export default function PolynomialConstruction(props) {
                 </div>
                 {' '}
                 <button
-                    className="btn btn-default"
+                    className="btn btn-default btn-construct"
                     onClick={props.onConstruction}
                 >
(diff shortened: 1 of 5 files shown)


scanomatic/ui_server_data/js/ccc/components/PolynomialConstruction.js
```diff
@@ -25,21 +25,40 @@ export default function PolynomialConstruction(props) {
         );
     }
 
+    const degrees = ['1', '2', '3', '4', '5'];
```
local-minimum: skip first degree since that will not really work with the solver. (unless you updated that one too).

scanomatic/ui_server_data/js/ccc/components/PolynomialConstruction.js
```diff
@@ -25,21 +25,40 @@ export default function PolynomialConstruction(props) {
...
+                    <label>Degree of polynomial</label>
+                    <select
+                        className="degree form-control"
+                        onChange={props.onDegreeOfPolynomialChange}
+                        value={props.degreeOfPolynomial}
+                    >
+                        {degrees.map(d => <option key={d} value={d}>{d}</option>)}
+                    </select>
+                </div>
+                {' '}
```
local-minimum: What is this and why is it needed? Seems better to style a spacing if that is the problem.

scanomatic/ui_server_data/js/specs/components/PolynomialConstruction.spec.js
```diff
@@ -6,9 +6,12 @@ import PolynomialConstruction from
...
 
 describe('<PolynomialConstruction />', () => {
+    const onDegreeOfPolynomialChange = jasmine.createSpy('onDegreeOfPolynomialChange');
     const props = {
+        degreeOfPolynomial: 3,
         onConstruction: jasmine.createSpy('onConstruction'),
         onClearError: () => {},
+        onDegreeOfPolynomialChange,
         polynomial: {
             power: -7,
```
local-minimum: I have a knack for mixing `power` and `degree` as if they were the same word. That may not be very good. But I also wonder if we're starting (mostly my fault) make things confusing.

The `degreeOfPolynomial` refers to how the next polynomial will be constructed, while `polynomial.power` is a convenience field for the degree of the polynomial defined by `polynomial.coefficients`.

At least, it would probably be good here if the test data was updated to have `polynomial.power` equal `2`, which is what the coefficients reflect.

scanomatic/ui_server_data/js/ccc/components/PolynomialConstruction.js
gdetrez: No, I haven't updated anything on the python side. I'll remove 1

scanomatic/ui_server_data/js/ccc/components/PolynomialConstruction.js
gdetrez: This is a space character. It is used, apparently, in bootstrap's inline forms to space stuff.
It does look a bit awkward. On the other hand, I can't find of a good reason why using CSS margin would necessarily be better (assuming that's what you mean by "style a spacing").

scanomatic/ui_server_data/js/specs/components/PolynomialConstruction.spec.js
gdetrez: `polynomial.power` wasn't used so I just removed it.

PR #231, opened 2017-11-27
Add integration test for analysis_effector

gdetrez:
The test runs `analysis_effector` on a project with one image that has
been "compiled". The result is compared with the result produced by
running the analysis through the web ui.

This also fixes two bugs that prevented `analysis_effector` to run.


scanomatic/image_analysis/test/test_grid_array.py
```diff
@@ -53,6 +56,11 @@ def test_get_im_slicei_no_im(grid_cell, expected_type):
     (MockedGridCell(xy1=(1,), xy2=(1, 1)), NoneType, None),
     (MockedGridCell(xy1=(1, 1), xy2=(1, 1)), ndarray, (0, 0)),
     (MockedGridCell(xy1=(0, 1), xy2=(10, 20)), ndarray, (10, 19)),
+    (
```
local-minimum: Isn't this just duplicating the test-case on line 58?

gdetrez: No. First one uses a tuple the other a numpy array.

gdetrez:
> There seems to be a few files in the analysis folder that seems not to be used. Is there a reason for including them? Some of them are quite large.

No, I guess I can remove them.

> Side note, on same topic. The practice of placing the tests next to the code, in this case also makes our docker-image much larger than it needs to be. I wonder if that's something we should consider.

I'm not a huge fan of it either.

PR #233, opened 2017-11-29
Improving calibration poly guessing and tests of related functions

local-minimum:
It isn't physically reasonable that any of the coefficients may become negative, so the solver should not allow for this. Unfortunately I didn't find any way to limit search space for the coefficients in the `optimize.leastsq` so I did manual assurance making both the poly and the results are always positive. Reason it is needed in both is that the solver may overstep into negative coeffs, thinking it is actually using them an thus reporting them as negative while they were actually positive. 

Another way to do the same thing would have been to optimize not on the coefficients but the log of the coefficients, but I think that code may turn a bit more obscure.

I added tests also for surrounding methods that didn't have particular unit tests.

### DIFF
--- a/scanomatic/data_processing/calibration.py
+++ b/scanomatic/data_processing/calibration.py
@@ -589,8 +589,8 @@ def get_calibration_optimization_function(degree=5):
     arr = np.zeros((degree + 1,), np.float)
 
     def poly(data_store, c1, cn):
-        arr[-2] = c1
-        arr[0] = cn
+        arr[-2] = max(c1, 0)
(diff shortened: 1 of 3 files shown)


skymandr:
absing should work, but it's a bit of a hack. No more so than what you're currently doing however. =)

local-minimum:
I updated in what you are looking at so it abses, less hackish would be to have the function do:

```
np.exp(c1) * x ...
```

skymandr:
I wrote too soon, based on what you wrote in the original email. Looks good to me.

PR #234, opened 2017-11-30
CCC colony histograms

local-minimum:
A histogram a colony:

image


skymandr:
How is it decided which histograms to show?

local-minimum:
It shows them all (I'm btw trying to add a better svg test with help from @gdetrez

local-minimum:
Actually while it was possible to test, without the proper styling the tests show not very useful images...

scanomatic/data_processing/test/test_calibration.py
```diff
@@ -751,3 +751,54 @@ def test_calculate_polynomial(self):
...
+        assert colonies['source_values'][0] == [
+            0.21744831955999988, 1.2174483195599999, 2.2174483195599999,
+            3.2174483195599999, 4.2174483195599999, 5.2174483195599999,
+            7.2174483195599999, 9.217448319559999, 13.217448319559999,
+        ]
+        assert colonies['source_value_counts'][0] == [
+            11620, 10491, 10, 4, 1, 1, 1, 1, 2,
+        ]
+        assert colonies['target_values'][0] == 210000.0
+        assert colonies['source_values'][-1] == [
```
skymandr: couldn't this be more compactly written as `np.arange(4, 31) + .68582517095`? (same for line 781)

skymandr: (I realise that this would mean also changing the .ccc file, but since the data is obviously bogus anyway, it can't really matter that much to remove the trailing `00003`, right?

local-minimum: No I suppose it could be compacted, but I don't know if that actually is good idea since these are the actual data in the fixture (file `scanomatic/data_processing/test/data/test_good.ccc`) and to me it is more transparent that the test is like it is. The ccc-file is real measurements too. Don't know if there's a point to it, but they are.

local-minimum:
I mean one could arbitrary group them in sets of two-three, but since the data will tend to be very similar, I think this would be very unhelpful to the user.
Another option would be to have a global histogram for all 96 colonies to show distribution of darkness in general in the data-set. That however, I think would be a separate PR.

skymandr:
What does your UX gut feeling tell you? Is your solution acceptable?

PR #240, opened 2017-12-08
Estimate all the coeffs (almost)

local-minimum:
Using all but x0.


local-minimum:
I think the coveralls warning is not related to this PR but that since the origin of this branch coverage has gone down in general due to moving of the tests.

@skymandr I'd like you to specifically look at the maths of this pr. I think it is valid to require x0 to be 0 in the solution or else we will require colony-detection not to be greedy and include pixels surrounding the colony. And that seems like a bad thing to require.

scanomatic/data_processing/calibration.py
```diff
@@ -586,13 +586,13 @@ def _collect_all_included_data(ccc):
...
 def get_calibration_optimization_function(degree=5):
 
-    arr = np.zeros((degree + 1,), np.float)
+    coeffs = np.zeros((degree + 1,), np.float)
 
-    def poly(data_store, c1, cn):
-        arr[-2] = abs(c1)
-        arr[0] = abs(cn)
+    def poly(data_store, *guess):
+        coeffs[:] = np.abs(guess)
```
skymandr: Use `exp` to avoid 0 attractor, but maybe in different PR.

scanomatic/data_processing/calibration.py
```diff
@@ -586,13 +586,13 @@ def _collect_all_included_data(ccc):
...
+        coeffs[-1] = 0
```
skymandr: I think it would be cleaner, and maybe give `leastsq` less of a headache, to exclude the const coeff from the guess and insert it in the call to `polyval` e.g. by `np.r_[coeffs, 0]` or similar.

tests/unit/data_processing/test_calibration.py
```diff
@@ -68,23 +68,9 @@ def test_returns_expected_values(self):
...
-        data = calibration.CalibrationData(
-            source_value_counts=[[10, 2], [3]],
-            source_values=[[1, 2], [3]],
-            target_value=[100, 126]
-        )
-        c1 = -2
-        c2 = -4
-        sums = colony_summer(data, c1, c2)
+        c1 = np.log(2)  # e^x = 2
+        c2 = np.log(4)  # e^x- =4
```
skymandr: Typo in comment (spurious `-` and in inconsistent white space compared to lines 52 and 53). Either fix or remove (if you really feel a comment is needed, please make it more verbose).

tests/unit/data_processing/test_calibration.py
```diff
@@ -703,54 +683,119 @@ def test_too_little_data_raises(self, data_store):
...
+            (8,),
+        ),
+        (
+            calibration.CalibrationData([[2]], [[1]], []),
+            (-99, -99, 0, -99),
+            (4,),
+        ),
+        (
+            calibration.CalibrationData([[2]], [[1]], []),
+            (-99, -99, -99, 0),
```
skymandr: The `-99` is a little opaque. Perhaps replace with a global constant (e.g. `INFINITY`, or `INFTY` or `INF`) and reuse to make clearer what the desired effect is?

PR #242, opened 2017-12-11
Remove xml export

local-minimum:
XML exports haven't been used for a couple of years, so it's time to remove them.

I did need to make the reading of serialized `analysis.model` files possible from before the deprecation and therefore made the matching of if files contains serializations of the models of interest more permissive. Added some tests of this change in the terms of `AnalysisModelFactory`.


tests/unit/models/factories/data/analysis.current.model
```diff
@@ -0,0 +1,173 @@
+[XML]
```
gdetrez: Leftover XML stuff?

tests/unit/models/factories/test_analysis_factory.py
```diff
@@ -33,7 +40,7 @@ def test_model_has_ccc_id(self, analysis_model):
     def test_model_can_serialize(self, analysis_model):
 
         serial = AnalysisModelFactory.serializer.serialize(analysis_model)
-        assert len(serial) == 3
+        assert len(serial) == 2
```
gdetrez: What does this has to do with XML?

scanomatic/generics/abstract_model_factory.py
```diff
@@ -196,9 +197,18 @@ def create(cls, **settings):
         return cls.MODEL(**settings)
 
     @classmethod
-    def all_keys_valid(cls, keys):
-
-        return set(tuple(cls.default_model.keys())).issuperset(keys)
+    def matching_key_set(cls, keys):
```
gdetrez: This method is strange. As far as I understand it checks that the keys in the serialized data are "75% correct", for any "75%"?
If, as I understand, the point is to check update the `all_keys_valid` to ignore the deprecated key related to XML I think a better approach would be to explicitly ignore those keys keys with something like:
`set(keys) - {'deprecated1', 'deprecated2', ...}`

scanomatic/generics/abstract_model_factory.py
```diff
@@ -196,9 +197,18 @@ def create(cls, **settings):
...
+
+        expected = set(tuple(cls.default_model.keys()))
```
gdetrez: Why `tuple`?

scanomatic/generics/abstract_model_factory.py
```diff
@@ -196,9 +197,18 @@ def create(cls, **settings):
...
+        common = len(expected.intersection(keys))
+        diff = len(expected.difference(keys))
+        n_expected = len(expected)
+        n_keys = len(keys)
+        return common > diff and (
+            common == n_expected or
+            common == n_keys or
+            common / float(n_expected) > .75 and common / float(n_keys) > .75
```
gdetrez: Remove `float`?

tests/unit/models/factories/test_analysis_factory.py
local-minimum: Serialization puts nested models in sequence,  so without the XML modell it should be one less

scanomatic/generics/abstract_model_factory.py
local-minimum: fixed

scanomatic/generics/abstract_model_factory.py
local-minimum: fixed

scanomatic/generics/abstract_model_factory.py
local-minimum: in principle yes but then it shouldn't be here but on that specific class.

tests/unit/models/factories/data/analysis.current.model
local-minimum: Fixed

scanomatic/generics/abstract_model_factory.py
local-minimum: Alright I fixed this by putting the specific allowances on the specific factory instead. The strictness was way too allowing to be really meaningful. I'm a bit scared that we might now disallow old, but otherwise usable model-serializations for other models. I don't really want to extend this PR with those tests too though.

local-minimum:
Feel free to merge this next week if you think it is fine :)

local-minimum:
@gdetrez Ping

PR #243, opened 2017-12-11
Image upload endpoint

gdetrez:
Add a new endpoint `/api/scans` to post scans (images). The scans are saved in the project directory according to the current file pattern.

I couldn't identify a clear pattern in the design of the current API endpoints so I designed the new endpoint as a REST API because we are already using it at molflow for other projects and there is plenty of resources and tools for it. I suggest that future endpoint should also use REST.

There is already code in Scan-o-Matic that saves scan images. I did try to refactor this to use the same code as the new endpoint but the image-saving code is quite entangled in the scanning "effector" which proved quite painful to get under tests. Or understand.

### DIFF
--- a/scanomatic/io/scanstore.py
+++ b/scanomatic/io/scanstore.py
@@ -1,0 +1,27 @@
+import os
+from scanomatic.io.paths import Paths
+
+
+class UnknownProjectError(Exception):
+    pass
+
+
+class ScanStore(object):
+    def __init__(self, path):
+        self._path = path
+
+    def add_scan(self, project_fullname, scan):
+        project_dir = os.path.join(self._path, project_fullname)
+        if not os.path.isdir(project_dir):
+            raise UnknownProjectError()
+        project_name = os.path.basename(project_fullname)
+        filename = Paths().experiment_scan_image_pattern.format(
+            project_name,
+            '{:04d}'.format(scan.index),
+            scan.timedelta.total_seconds(),
+        )
+        path = os.path.join(self._path, project_fullname, filename)
+        with open(path, 'w') as f:
+            f.write(scan.image.read())
+            f.flush()
+            os.fsync(f)
--- a/tests/system/conftest.py
+++ b/tests/system/conftest.py
@@ -2,10 +2,15 @@ import pytest
 import requests
 from selenium import webdriver
 
+import py.path
+
 
 @pytest.fixture(scope='session')
 def docker_compose_file(pytestconfig):
-    return pytestconfig.rootdir.join('docker-compose.yml')
+    return [
+        pytestconfig.rootdir.join('docker-compose.yml'),
+        py.path.local(__file__).dirpath().join('docker-compose.override.yml'),
+    ]
 
 
 @pytest.fixture(scope='session')
--- a/tests/system/test_scanner_api.py
+++ b/tests/system/test_scanner_api.py
@@ -1,0 +1,59 @@
+from StringIO import StringIO
+from collections import namedtuple
+from hashlib import sha256
+import httplib
+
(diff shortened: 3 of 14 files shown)


scanomatic/io/scanstore.py
```diff
@@ -0,0 +1,27 @@
...
+    pass
+
+
+class ScanStore(object):
+    def __init__(self, path):
+        self._path = path
+
+    def add_scan(self, project_fullname, scan):
+        project_dir = os.path.join(self._path, project_fullname)
+        if not os.path.isdir(project_dir):
```
local-minimum: You see us making another api-endpoint for creating a project too then? Seems reasonable. Just wanted to be sure.

tests/system/conftest.py
```diff
@@ -2,10 +2,15 @@
...
 
+import py.path
+
 
 @pytest.fixture(scope='session')
 def docker_compose_file(pytestconfig):
-    return pytestconfig.rootdir.join('docker-compose.yml')
+    return [
+        pytestconfig.rootdir.join('docker-compose.yml'),
+        py.path.local(__file__).dirpath().join('docker-compose.override.yml'),
```
local-minimum: Just as a question, is it a benefit versus `os.path.dirname(__file__)`?

tests/system/test_scanner_api.py
```diff
@@ -0,0 +1,59 @@
...
+
+@pytest.fixture
+def image():
+    data = b'I am an image'
+    return namedtuple('image', ['image', 'filename', 'digest'])(
+        StringIO(data), 'my_image.tiff', sha256(data).hexdigest()
+    )
+
+
+def get_nb_of_images(scanomatic, project):
```
local-minimum: typo? `nb` => `nr`

tests/system/test_scanner_api.py
```diff
@@ -0,0 +1,59 @@
...
+    response = requests.get(
+        scanomatic + '/api/tools/path/root/{}'.format(project),
+        params={'suffix': '.tiff', 'isDirectory': 0, 'checkHasAnalysis': 0},
```
local-minimum: To be sure this works, the tests needs to ensure that there's nothing in that directory before it starts, because if there are other tests that create folders in the same directory the count will be wrong.

tests/unit/ui_server/test_experiment_api.py
```diff
@@ -88,3 +106,61 @@ def test_dont_keep_previous_qc(self, validator_mock, test_app):
         })
         assert response.status_code == 200
         assert json.loads(response.data)['success']
+
+
+class TestPostScan:
```
local-minimum: I would like a test with a bad `digest`.

gdetrez: There's a unit test

scanomatic/io/scanstore.py
gdetrez: Maybe.
In retrospect I actually wish we had gone through the whole process before picking this task...

tests/system/conftest.py
gdetrez: Not really. Just a bit nicer API.

tests/system/test_scanner_api.py
gdetrez: Interesting. *nb* is probably a "[frenshism](https://fr.wiktionary.org/wiki/Nb.)" but as far as I can tell *nr* might be a "[swedishism](https://en.wiktionary.org/wiki/nr.#Swedish)" :stuck_out_tongue_winking_eye:.
(The recommended English abbreviation being *no*.)
I think I'll just spell it out...

PR #246, opened 2018-01-02
Reactivate ccc test

local-minimum:
We had a skipped ccc-test due to lacking real and good data.
Made such data, reactivated the test.
To make the test work well I also needed to make sure colonies came in a good order so I made them sort by measured size (will affect the histogram sort order on ccc too now).


scanomatic/data_processing/calibration.py
```diff
@@ -714,32 +726,25 @@ def construct_polynomial(identifier, power):
...
-    modes = [
-        values[np.argmax(counts)] for values, counts in
-        zip(data_store.source_values, data_store.source_value_counts)
-    ]
-    sort_order = np.argsort(modes)
-    values = [data_store.source_values[colony] for colony in sort_order]
-    counts = [data_store.source_value_counts[colony] for colony in sort_order]
-    value_sorts = [np.argsort(vector) for vector in values]
+    if data_store.source_values:
+        min_values = min(min(vector) for vector in data_store.source_values)
```
skymandr: This is actually a scalar, so the plural s is perhaps misleading. No biggie though.

tests/unit/data_processing/test_calibration.py
```diff
@@ -129,6 +129,12 @@ def edit_bad_slope_ccc():
     calibration.reload_cccs()
 
 
+@pytest.fixture(scope='function')
+def full_ccc():
+    _ccc = _fixture_load_ccc('data/TESTUMz.ccc')
```
skymandr: `TESTUMz`? A pretty opaque name...

local-minimum: :grin: Was the name generated by the ccc. Figured it might be useful to use actual naming conventions of the files as they are made by the CCC, Maybe not? I can change if you think not.

scanomatic/data_processing/calibration.py
local-minimum: Yes but and I wasn't sure what I thought best. `values` and `counts` makes the names makes the connection to the `source_values` and `source_value_counts` a bit clearer. Can change if you want to.

skymandr: Yeah, I wasn't sure either. Your call.

tests/unit/data_processing/test_calibration.py
skymandr: It's probably fine if that's what the CCC produces, though why it uses that naming convention I don't know..?

PR #247, opened 2018-01-04
New year's cleanin'

local-minimum:
Key features:

* external js now resides in own folder and has own endpoint
* all endpoints that deliver pages are now in `ui_pages.py`
* in `ui_server.py` now only remains resource-endpoints.
* no more mixed api/page serving endpoints.
* cleaning up the experiment api
* system tests that checks that pages and resources can be accessed.
* fixed failing tests due to new api-endpoints
* removed some unused stuff
* fixed some lint errors.

**Note:** While this does not really introduce any new lint errors, there are a million long lines that appear as changed just because code moved around or lost a bit of indentation. These will not be fixed in this PR.


scanomatic/ui_server_data/qc_norm.html
```diff
@@ -3,18 +3,19 @@
     <head>
         <title>Quality Control Prototype</title>
         <meta charset="utf-8" />
-        <script src="js/jquery-2.2.3.js"></script>
-        <script src="js/bootstrap.js"></script>
-        <script src="js/d3.js"></script>
+        <script src="js/external/jquery.js"></script>
+        <script src="js/external/bootstrap.js"></script>
+        <script src="js/external/d3.js"></script>
```
skymandr: Where do we use d3? I thought you opted for another library? Or does that one depend on d3? General question: are we using minimised versions of externals when we can?

scanomatic/ui_server_data/maintain.html
```diff
@@ -20,61 +19,9 @@ <h2 id="version-info">Checking version...</h2>
             <p><em>
                 If Scan-o-Matic currently have jobs it is working on or in
                 queue, it is generally not recommended to do anything to the
-                server.  In most cases it still should be alright, but some
-                cases remain to be fully tested and verified (especially
-                rebooting server while scanning projects are running).
-                Shutting down the server while scanning experiments are running
-                will always cause issues as the server is needed to acquire
-                scanners.
+                server. You will terminate all.
```
skymandr: all what?

scanomatic/ui_server_data/qc_norm.html
local-minimum: I think qc uses it still. Or somewhere. It was used anyhow

scanomatic/ui_server_data/js/analysis.js
```diff
@@ -136,24 +136,24 @@ function get_regridding_setting(i) {
...
     path = get_dir();
     get_path_suggestions(
         input,
         true,
-        "",
+        '',
         null,
         function(data, status) {
 
             //TODO: For some reason popup don't appear...
```
skymandr: Still an issue?

scanomatic/ui_server/management_api.py
```diff
@@ -1,187 +1,65 @@
...
-                os.kill(proc_id, signal.SIGKILL)
-                server_ids.add(proc_id)
-
-            if rpc_client.online:
-                return jsonify(success=False,
-                               reason="Tried to kill processes {0}, but somehow server is online".format(server_ids))
-
-            return jsonify(success=True)
-
     @app.route("/api/app/<action>", methods=['post', 'get'])
```
skymandr: This endpoint seems completely superfluous now. Why not just `/api/app/version`?

scanomatic/ui_server_data/js/analysis.js
local-minimum: Sometimes you have to type stuff for it to start to appear, no real point in working on it, rather fix when upping to react or so.

scanomatic/__init__.py
```diff
@@ -1,19 +1,18 @@
 #!/usr/bin/env python
 """Part of analysis work-flow that holds a grid arrays"""
+import os
 
 __author__ = "Martin Zackrisson"
 __copyright__ = "Swedish copyright laws apply"
 __credits__ = ["Martin Zackrisson", "Mats Kvarnstroem", "Andreas Skyman", ""]
```
skymandr: And Gregoire and Fredrik?

scanomatic/__init__.py
```diff
@@ -1,19 +1,18 @@
...
 __license__ = "GPL v3.0"
-__version__ = "v2.0.4"
+__version__ = "v2.1.6"
 __maintainer__ = "Martin Zackrisson"
 __email__ = "martin.zackrisson@gu.se"
```
skymandr: No longer valid?

scanomatic/ui_server_data/maintain.html
local-minimum: The world :)  Fixed...

scanomatic/ui_server_data/qc_norm.html
skymandr: ok

scanomatic/ui_server/management_api.py
local-minimum: agree, fixed.

scanomatic/__init__.py
local-minimum: True that. Perhaps @gdetrez and @dindinoh could PR on it if they want in on the credz.

scanomatic/ui_server_data/js/analysis.js
skymandr: I agree. Just wondered if perhaps it had been solved as an unforseen side effect! =)

I don't suggest you do it, but do you think an xfailing test could be written for this?

scanomatic/__init__.py
local-minimum: Version changes aren't really valid but they show in the app and ensure you don't load outdated js/css/fonts/html so I don't think it hurts upping numbers and keeping them.

scanomatic/ui_server_data/js/analysis.js
local-minimum: It's at system-test level basically... not sure it is worth time it takes to write it.

scanomatic/__init__.py
skymandr: I meant your email.

scanomatic/ui_server_data/js/analysis.js
skymandr: Ok. Perhaps make an issue for it if there isn't one?

local-minimum: I'll make a task for what would invalidate the issue instead.

scanomatic/__init__.py
local-minimum: true, don't want mails though :grinning:  
will remove in next PR instead

local-minimum:
Should add I've clicked about and tried doing a lot of the stuff in the UI.

scanomatic/__init__.py
skymandr: Ok.

scanomatic/__init__.py
skymandr: Ok.

scanomatic/ui_server/management_api.py
skymandr: Looks good to me now.

scanomatic/ui_server_data/js/analysis.js
skymandr: Good.

PR #248, opened 2018-01-04
Remove stuff

local-minimum:
Since this is based on #247, I recommend waiting wit review until that one is merged.
This removes a lot of stuff, but should be fairly easy to follow after that other is merged.

Removes modules (both js and py), removes some scripts. Removes some functions that were useless or not used.


scanomatic/ui_server_data/js/analysis.js
```diff
@@ -2,7 +2,7 @@ var gridplates = null;
 var localFixture = true;
 var path = '';
 
-function toggleLocalFixture(caller) {
+function analysisToggleLocalFixture(caller) {
```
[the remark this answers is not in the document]
local-minimum: It is used, by the html

scanomatic/ui_server_data/js/compile.js
```diff
@@ -120,7 +120,7 @@ function setOnAllImages(included) {
 }
 
 
-function toggleLocalFixture(caller) {
+function compileToggleLocalFixture(caller) {
```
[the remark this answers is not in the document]
local-minimum: It is used, by the html

scanomatic/ui_server_data/js/analysis.js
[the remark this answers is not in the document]
local-minimum: Is used by the html

scanomatic/ui_server_data/js/compile.js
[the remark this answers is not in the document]
local-minimum: Is used by the html

scanomatic/__init__.py
```diff
@@ -6,9 +6,8 @@
 __copyright__ = "Swedish copyright laws apply"
 __credits__ = ["Martin Zackrisson", "Mats Kvarnstroem", "Andreas Skyman", ""]
 __license__ = "GPL v3.0"
-__version__ = "v2.1.6"
+__version__ = "v2.1.7"
```
skymandr: What's the bump policy on this?

local-minimum: `[0-9].[0-9].[0 - 999]` last digit can be increased per commit if one pleases. Middle some sort of relevant change, first major release, think SoM as a service would be 3.0.0 or so, v2.2.0 might be the first that drops the ability to scan by itself. Adding CCC ior other substantial features definitely a new middle.

scanomatic/ui_server_data/js/fixtures.js
```diff
@@ -296,10 +287,6 @@ function testAsGrayScale(plate) {
     }
 }
 
-function OnEnterFixtureName() {
-    SetAllowDetect();
-}
-
 function SetAllowDetect() {
```
skymandr: Is this tested?

scanomatic/__init__.py
skymandr: I find these guidelines very helpful: https://semver.org/

scanomatic/ui_server_data/js/fixtures.js
local-minimum: There's system-test for creating fixtures in the UI no. But I tested it now. It works as intended.

scanomatic/__init__.py
local-minimum: Kind of like that but not strict. Inspired by.

skymandr: I think that's a discussion worth having. Since the version system looks like semantic versioning, some users may expect a minor bump to be backwards compatible, which is not the case for the removal of scanning for instance, so maybe that should rather be a major. By the similar argument, this PR should perhaps bump the minor rather than the third.

skymandr: (I think this might also be what pip etc. assume, so it's worth thinking about what we mean and how it will be interpreted.)

local-minimum: Discussion sure, I'm 100% non of current users know anything about this and just expect it to increment... I can change, but major release would be semantically wrong in the context of its own version history.

local-minimum: But discussion shouldn't be here though.

PR #252, opened 2018-01-11
Rename folder `ccc` to `src`

local-minimum:
Since we're getting more stuff that needs compiling and we want to reuse components and such this moves all the ccc-files and changes some names were needed.


webpack.config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
     entry: {
-        ccc: ['./scanomatic/ui_server_data/js/ccc/index.jsx'],
+        ccc: ['./scanomatic/ui_server_data/js/src/ccc_index.jsx'],
```
gdetrez: Maybe just `ccc.js`?

scanomatic/ui_server_data/js/src/api.js
```diff
@@ -41,6 +39,8 @@ class API {
     }
 }
 
+export const HasJquery = () => !!$;
```
gdetrez: Remove

local-minimum: fixed

webpack.config.js
local-minimum: fixed

PR #263, opened 2018-01-22
Add "Z" to ScanJob start datetime serialization

gdetrez:
- Update the serialization datetime in the API to be in the format
  YYYY-MM-DDThh:mm:ssZ (This makes it easier to parse)
- Add checks to enforce a ScanJob start value to have UTC timezone.


scanomatic/ui_server/serialization.py
```diff
@@ -7,5 +11,6 @@ def job2json(job):
         'scannerId': job.scanner_id,
     }
     if job.start is not None:
-        obj['start'] = job.start.isoformat()
+        assert is_utc(job.start)
```
local-minimum: I think this change requires the endpoints that use it to catch the exception and return a proper `json_abort` with a reason. Right?

gdetrez: No. The assertion is there to enforce/document the assumption made in the next line (the formatting is only correct if the datetime is UTC). I think it isn't necessary to catch it the same way that it isn't necessary to catch `AttributeError` even though it would be raised if `start` wasn't a `datetime`.

local-minimum: But will it not happen when start is `None` then? And that endpoint not able to return jobs that haven't started yet. Maybe I'm confused though...

local-minimum: strike that hehe

PR #264, opened 2018-01-22
Add `/start` endpoint for scan jobs

gdetrez:
Add the following endpoints:
- `GET /api/scan-jobs/$id`: get a single scanjob by id
- `POST /api/scan-jobs/$id/start`: start a scanjob

In REST slang the first would be a document resource and the second a controller resource.

I'm a bit unhappy with the "start" controller having the same name as the "start" (time) field but I'm not sure it's really a problem. I can rename the field if someone thing it's going to cause issues.

This somewhat encroaches on a future task (start job when button is clicked), but I need it to write tests using the scanjob API.

### DIFF
--- a/scanomatic/ui_server/scan_jobs_api.py
+++ b/scanomatic/ui_server/scan_jobs_api.py
@@ -5,6 +5,7 @@ from uuid import uuid1
 
 from flask import request, jsonify, Blueprint, current_app
 from flask_restful import Api, Resource
+import pytz
 from werkzeug.exceptions import NotFound
 
 from scanomatic.io.scanning_store import (
@@ -103,11 +104,12 @@ class ScanJobStartController(Resource):
             job = db.get_scanjob(scanjobid)
         except ScanJobUnknownError:
             raise NotFound
+        now = datetime.now(pytz.utc)
         if job.start is not None:
             return json_abort(CONFLICT, reason='Scanning Job already started')
-        if db.get_current_scanjob(job.scanner_id, datetime.now()) is not None:
+        if db.get_current_scanjob(job.scanner_id, now) is not None:
             return json_abort(CONFLICT, reason='Scanner busy')
-        db.update_scanjob(job._replace(start=datetime.now()))
+        db.update_scanjob(job._replace(start=now))
         return '', OK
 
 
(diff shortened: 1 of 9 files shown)


scanomatic/ui_server/scan_jobs_api.py
```diff
@@ -80,3 +85,34 @@ def scan_jobs_list():
...
+    def post(self, scanjobid):
+        db = current_app.config['scanning_store']
+        try:
+            job = db.get_scanjob(scanjobid)
+        except ScanJobUnknownError:
+            raise NotFound
+        now = datetime.now(pytz.utc)
+        if job.start is not None:
+            return json_abort(CONFLICT, reason='Scanning Job already started')
+        if db.get_current_scanjob(job.scanner_id, now) is not None:
```
local-minimum: This feels an obscure way to ask the qustions. Feels like it should be `db.get_scanner_status(job.scanner_id, now)` or similar.

scanomatic/ui_server/scan_jobs_api.py
```diff
@@ -80,3 +85,34 @@ def scan_jobs_list():
...
+            return json_abort(CONFLICT, reason='Scanner busy')
+        db.update_scanjob(job._replace(start=now))
+        return '', OK
+
+
+api = Api(blueprint)
+api.add_resource(ScanJobStartController, '/<scanjobid>/start')
```
local-minimum: alternative could be something like `/<scanjobid>/set-start` but that also sound bit strange.

gdetrez: I would rather change the field name.

local-minimum: do that: `start_time` ? If so need to update py2json conversion too so becomes `startTime`

scanomatic/ui_server/scan_jobs_api.py
gdetrez: That's a bit too vague. I can add `db.has_current_job(scannerid, now)` if you'd prefer.

local-minimum: yes sound :1st_place_medal:

scanomatic/ui_server/scan_jobs_api.py
gdetrez: Fixed

scanomatic/ui_server/scan_jobs_api.py
gdetrez: Fixed

PR #266, opened 2018-01-24
Fix an error in the endpoint to get a scanner job

gdetrez:
This also converts the tests to integration tests.

### DIFF
--- a/tests/integration/api/test_scanners.py
+++ b/tests/integration/api/test_scanners.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-from httplib import OK, NOT_FOUND, CREATED
+from httplib import OK, NOT_FOUND
 import json
 import uuid
 
--- a/tests/integration/api/test_scanners.py
+++ b/tests/integration/api/test_scanners.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-from httplib import OK, NOT_FOUND
+from httplib import OK, NOT_FOUND, CREATED
 import json
 import uuid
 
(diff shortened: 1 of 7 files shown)


tests/integration/api/test_scanners.py
```diff
@@ -0,0 +1,75 @@
...
+    app.register_blueprint(scan_jobs_api.blueprint, url_prefix="/scan-jobs")
+    add_configs(app)
+    return app
+
+
+@pytest.fixture
+def apiclient(client):
+    class APIClient:
+        def create_scanning_job(
+            self, scannerid, name=None, duration=600, interval=300,
```
local-minimum: This might be out of scope, but it feels like the endpoint should not allow None or empty string names.

tests/integration/api/test_scanners.py
```diff
@@ -0,0 +1,75 @@
...
+        assert response.json is None
+
+    def test_has_scanjob(self, apiclient):
+        jobid = apiclient.create_scanning_job(self.SCANNERID).json['jobId']
+        with freeze_time('1985-10-26 01:20', tz_offset=0):
+            apiclient.start_scanning_job(jobid)
+        with freeze_time('1985-10-26 01:21', tz_offset=0):
+            response = apiclient.get_scanner_job(self.SCANNERID)
+        assert response.status_code == OK
+        assert response.json['identifier'] == jobid
```
local-minimum: should it not test the `name` too?

tests/integration/api/test_scanners.py
gdetrez: I think it doesn't (you wrote it, you should know :stuck_out_tongue_winking_eye:)

tests/integration/api/test_scanners.py
gdetrez: Why not I guess. Updated

PR #270, opened 2018-01-29
Automatic fixing of js-errors

local-minimum:
This would allow stickler-ci to do, or propose? automatic fixing of js-errors.

### DIFF
--- a/.stickler.yml
+++ b/.stickler.yml
@@ -3,7 +3,8 @@ linters:
   pep8:
     ignore: 'W503'
   py3k: {  }
-  eslint: {  }
+  eslint:
+      fixer: true
 files:
   ignore:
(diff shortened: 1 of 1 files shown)


gdetrez:
Can you make an other pull request based on this branch with a lint error to see what happens? (with the idea that this other pull request won't be merged)

.stickler.yml
```diff
@@ -1,9 +1,13 @@
...
+  enable: true
+
 linters:
   flake8: {  }
   pep8:
     ignore: 'W503'
   py3k: {  }
-  eslint: {  }
+  eslint:
+      fixer: true
```
gdetrez: Wrong indentation size (sort of ironic :laughing: )

PR #276, opened 2018-01-31
Get status tests

skymandr:
I've added tests for the `scanning_store` methods, but I suspect that there are tests missing for the API endpoint. I'll have a look first myself, but please tell me if you spot anything obvious that I've missed.

### DIFF
--- a/tests/unit/io/test_scanning_store.py
+++ b/tests/unit/io/test_scanning_store.py
@@ -7,8 +7,8 @@ from mock import patch
 from pytz import utc
 
 from scanomatic.io.scanning_store import (
-    ScanningStore, ScanJobCollisionError, ScanJobUnknownError, Scanner,
-    DuplicateIdError, UnknownIdError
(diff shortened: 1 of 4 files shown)


tests/unit/io/test_scanning_store.py
```diff
@@ -76,6 +76,54 @@ def test_has_no_scanner_if_not_using_test_scanners(self):
...
+
+    def test_no_get_scanner_status_list_unknown(self, scanning_store):
+        with pytest.raises(UnknownIdError):
+            scanning_store.get_scanner_status_list("42")
+
+    def test_get_lastest_scanner_status(self, scanning_store):
+        assert scanning_store.get_latest_scanner_status(
+            '9a8486a6f9cb11e7ac660050b68338ac') is None
+
+    def test_add_scanner_status(self, scanning_store):
```
local-minimum: possibly could be nice to know that it takes the latest status. I.e. add one and then one and make sure you get the newest.

PR #283, opened 2018-02-05
Add "start time" and "images to send" to scanner status 

gdetrez:


scanomatic/ui_server/scanners_api.py
```diff
@@ -74,23 +74,29 @@ def _add_scanner(scanner_id):
...
+        dest='start_time',
+        type=inputs.datetime_from_iso8601,
+        required=True,
+    )
+    parser.add_argument(
+        'imagesToSend',
+        dest='images_to_send',
+        type=inputs.natural,
+        required=True,
+    )
```
local-minimum: Why not add `nextScheduledScan` too?

scanomatic/util/datetime.py
```diff
@@ -1,6 +1,11 @@
...
+from datetime import timedelta
+
 import pytz
 
 
 def is_utc(dt):
-    return dt.tzinfo == pytz.utc
+    return (
+        dt.tzinfo is not None
+        and dt.tzinfo.utcoffset(dt) == dt.tzinfo.dst(dt) == timedelta(0)
```
local-minimum: I'm not a big fan of `X == Y == Z` but maybe it's fine.

scanomatic/ui_server/scanners_api.py
gdetrez: Just forgot this one...

PR #284, opened 2018-02-05
Default pinning

local-minimum:
Make `(32, 48)` default pinning.
Also removed `self._scanning_instructions` as it wasn't used.


gdetrez:
The default pining format should probably be a constant somewhere sane instead of buried inside a method.
Beside that, I can't really say if this is correct or not, mabye @skymandr  has more experience with this code?

local-minimum:
@gdetrez I moved it to a better place, will attempt making an analysis using this, from compilation, having no pinning-instructions

local-minimum:
It worked! 🍰

PR #285, opened 2018-02-07
Compile scanjob

gdetrez:
- Add a `projectdirectory` arg to compile page 
- Add a link to compilation in the ScanningJobPanel
- Add a system test that creates a scanjob, post a scan, click the new "compile" link and checks the project path and that the images is detected.


skymandr:
That's a pretty heft coverage decrease... :-/

gdetrez:
Whoops. Should be fixed.

tox.ini
```diff
@@ -5,6 +5,7 @@ skipsdist = true
 [testenv]
 deps =
     -rrequirements.txt
+    blinker
```
local-minimum: What is this? Don't see it used.

gdetrez: It is used by flask signal extension mechanism which I used in the test to get the template context.

PR #286, opened 2018-02-09
Get scanner "owned" property from API

gdetrez:
- add function to get scanner job from API
- add helper function that get the list of scanner and their current
jobs from the API and builds a list of scanners with the `owned`
property as expected by the UI.

### DIFF
--- a/scanomatic/ui_server_data/js/src/helpers.js
+++ b/scanomatic/ui_server_data/js/src/helpers.js
@@ -226,3 +226,10 @@ export function uploadImage(ccc, file, fixture, token, progress) {
         })
         .then(() => imageId);
 }
+
+export function getScannersWithOwned() {
+    return API.getScanners().then(scanners =>
+        Promise.all(scanners.map(scanner =>
+            API.getScannerJob(scanner.identifier)
+                .then(job => Object.assign(scanner, { owned: job != null })))));
+}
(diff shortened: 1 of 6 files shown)


scanomatic/ui_server_data/js/src/helpers.js
```diff
@@ -226,3 +226,10 @@ export function uploadImage(ccc, file, fixture, token, progress) {
         })
         .then(() => imageId);
 }
+
+export function getScannersWithOwned() {
```
local-minimum: There's something surprising with the name, the `withOwned` bit but I have no idea how to rename the function to make it clearer.

scanomatic/ui_server_data/js/src/helpers.js
```diff
@@ -226,3 +226,10 @@ export function uploadImage(ccc, file, fixture, token, progress) {
...
+    return API.getScanners().then(scanners =>
+        Promise.all(scanners.map(scanner =>
+            API.getScannerJob(scanner.identifier)
+                .then(job => Object.assign(scanner, { owned: job != null })))));
```
local-minimum: I would expect `job` to be `undefined` and while the current implementation would evaluate correctly, it still injects a bit of a surprise. I think I would prefer `!!job` but I leave it to you to decide.

PR #288, opened 2018-02-13
Expose scanner metrics to prometheus

gdetrez:
Expose the following metrics to prometheus: `scanner_current_jobs`, `'scanner_queued_uploads`, `scanner_start_time_seconds`, `scanner_last_status_update_time_seconds`, `scanner_status_updates_total`. Each has a `scanner` label with the scanner id.

Some notes:
- Having a label with the scanner id sort-of goes againts prometheus instrumentation best practices as it's an unbounded dimension, at least in theory. But since those are important, it might be worth the extra disk usage. And it's not like we'll have millions of scanners...
- For somewhat similar reason, I added tests for the scanner metrics whereas we usually don't do that.
- The metrics don't expire: even if a scanner goes offline, scanomatic will expose the last value as long as it is not restarted. This means that prometheus built-in staleness detection will not work and one needs to look at an other metrics to see if the scanner is still alive (`scanner_last_status_update_time_seconds`). This mimics the behavior of the pushgateway. I'm not sure it's the best way thing to do in this case but it was the simplest to implement 😁

### DIFF
--- a/scanomatic/scanning/update_scanner_status.py
+++ b/scanomatic/scanning/update_scanner_status.py
@@ -64,11 +64,12 @@ def update_scanner_status(
         images_to_send=images_to_send,
     )
     db.add_scanner_status(scanner_id, status)
-    SCANNER_CURRENT_JOBS.labels(scanner=scanner_id).set(job is not None)
-    SCANNER_QUEUED_UPLOADS.labels(scanner=scanner_id).set(images_to_send)
-    SCANNER_START_TIME.labels(scanner=scanner_id).set(timestamp(start_time))
-    SCANNER_LAST_STATUS_UPDATE_TIME.labels(scanner=scanner_id).set_to_current_time()
-    SCANNER_STATUS_UPDATES.labels(scanner=scanner_id).inc()
+    labels = {'scanner': scanner_id}
+    SCANNER_CURRENT_JOBS.labels(**labels).set(job is not None)
+    SCANNER_QUEUED_UPLOADS.labels(**labels).set(images_to_send)
+    SCANNER_START_TIME.labels(**labels).set(timestamp(start_time))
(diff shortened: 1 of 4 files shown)


scanomatic/scanning/update_scanner_status.py
```diff
@@ -2,14 +2,43 @@
...
 from scanomatic.models.scanner import Scanner
 from scanomatic.models.scannerstatus import ScannerStatus
+from scanomatic.util.datetime import timestamp
 from scanomatic.util.generic_name import get_generic_name
 
 
+SCANNER_CURRENT_JOBS = Gauge(
+    'scanner_current_jobs',
+    'Number of job currently executed',
+    ['scanner']
```
local-minimum: Would it be of interest to have them by client first and then scanner? Client could be the api-host as an example.

tests/unit/scanning/test_update_scanner_status.py
```diff
@@ -0,0 +1,52 @@
...
+from prometheus_client import REGISTRY
+import pytest
+from pytz import utc
+
+from scanomatic.scanning.update_scanner_status import update_scanner_status
+
+
+@pytest.fixture
+def scanner():
+    return uuid4().hex
```
local-minimum: curiosity: is one or the other of the different uuid-versions preferred?

scanomatic/scanning/update_scanner_status.py
gdetrez: Prometheus already adds an instance label, that corresponds to the client.

tests/unit/scanning/test_update_scanner_status.py
gdetrez: I tend to use `uuid4` unless I have a good reason to: 

> If all you want is a unique ID, you should probably call uuid1() or uuid4(). Note that uuid1() may compromise privacy since it creates a UUID containing the computer’s network address. uuid4() creates a random UUID.

PR #290, opened 2018-02-15
server side of keepalive status scanomatic-scanning pr17

dindinoh:
https://github.com/Scan-o-Matic/scanomatic-scanner/pull/17

### DIFF
--- a/scanomatic/scanning/update_scanner_status.py
+++ b/scanomatic/scanning/update_scanner_status.py
@@ -37,6 +37,11 @@ SCANNER_STATUS_UPDATES = Counter(
     'Number of status updates',
     ['scanner']
 )
+SCANNER_CURRENT_DEVICES = Gauge(
+    'scanner_current_devices',
+    'Number of devices currently online',
+    ['scanner']
+)
 
 
 def UpdateScannerStatusError(Exception):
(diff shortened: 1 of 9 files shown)


scanomatic/scanning/update_scanner_status.py
```diff
@@ -70,6 +82,7 @@ def update_scanner_status(
     SCANNER_START_TIME.labels(**labels).set(timestamp(start_time))
     SCANNER_LAST_STATUS_UPDATE_TIME.labels(**labels).set_to_current_time()
     SCANNER_STATUS_UPDATES.labels(**labels).inc()
+    SCANNER_CURRENT_DEVICES.labels(**labels).set(devices is not None)
```
gdetrez: If I understand correctly, devices is a list so this is always going to be 1. Maybe you mean `len(devices)` instead? Also, maybe add this metrics to the existing tests.

scanomatic/ui_server/scanners_api.py
```diff
@@ -82,6 +82,12 @@ def scanner_status_update(scanner):
         type=inputs.natural,
         required=True,
     )
+    parser.add_argument(
+        'devices',
+        dest='devices',
+        action='append',
+        required=True,
```
gdetrez: I'm not sure but doesn't that mean that the request will be refused if there isn't at least one device? In any case, just to be safe, I think there should be a test with an empty list.

PR #292, opened 2018-02-16
Refactor ScanningStore interface

gdetrez:
Remove duplication and simplify ScanningStore:
- `add_x(somex)` ➡ `add(somex)`
- `get_x(id)` ➡ `get(X, id)`
- `get_all_x()` ➡ `find(X)`
- `update_x(somex)` ➡ `update(somex)`
- `has_x(...)` ➡ `exists(X, ...)`
- etc.

Remove `get_free_scanners` and corresponding API parameter.

Methods for more specific queries/updates still exists (e.g. `get_current_job`).

### DIFF
--- a/scanomatic/io/scanning_store.py
+++ b/scanomatic/io/scanning_store.py
@@ -1,17 +1,12 @@
 from __future__ import absolute_import
+from collections import defaultdict
 from datetime import datetime
-import os
 import pytz
 
+from scanomatic.models.scan import Scan
+from scanomatic.models.scanjob import ScanJob
 from scanomatic.models.scanner import Scanner
-
-
-class ScanJobCollisionError(ValueError):
-    pass
-
-
-class ScanJobUnknownError(ValueError):
-    pass
+from scanomatic.models.scannerstatus import ScannerStatus
 
 
 class DuplicateIdError(ValueError):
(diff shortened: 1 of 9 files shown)


scanomatic/io/scanning_store.py
```diff
@@ -23,133 +18,85 @@ class DuplicateNameError(ValueError):
...
 
-    def add_scanner(self, scanner):
-        if self.has_scanner(scanner.identifier):
-            raise DuplicateIdError(
-                "Cannot add duplicate scanner with id '{}'".format(
-                    scanner.identifier)
-            )
-        elif self.get_scanner_by_name(scanner.name) is not None:
+    def add(self, item):
+        store = self._get_store(type(item))
```
local-minimum: it feels like a whole bunch of `type(item)` in this method, perhaps store it as variable?

scanomatic/io/scanning_store.py
```diff
@@ -23,133 +18,85 @@ class DuplicateNameError(ValueError):
...
-    def remove_scanjob(self, identifier):
-        if identifier in self._scanjobs:
-            del self._scanjobs[identifier]
-        else:
-            raise ScanJobUnknownError(
-                "{} is not a known job".format(identifier)
-            )
-
-    def get_scanjob(self, identifier):
+    def get(self, klass, id_):
```
local-minimum: curiosity: why `klass` and not `class_` considering the use of `id_` ?

scanomatic/io/scanning_store.py
```diff
@@ -23,133 +18,85 @@ class DuplicateNameError(ValueError):
...
-    def add_scanjob(self, job):
-        if job.identifier in self._scanjobs:
-            raise ScanJobCollisionError(
-                "{} already used".format(job.identifier)
+                    item.name)
             )
-        self._scanjobs[job.identifier] = job
+        if type(item) is Scan and not self.exists(ScanJob, item.scanjob_id):
+            raise UnknownIdError(ScanJob, item.scanjob_id)
+        self._get_store(type(item))[item.identifier] = item
```
local-minimum: you already have the right `store` above on line 48.

scanomatic/io/scanning_store.py
```diff
@@ -23,133 +18,85 @@ class DuplicateNameError(ValueError):
...
+    def find(self, klass, **constraints):
+        store = self._get_store(klass)
+        for item in store.values():
+            for key in constraints:
+                if getattr(item, key) != constraints[key]:
+                    break
+            else:
+                yield item
+
+    def exists(self, klass, identifier=None, **constraints):
```
local-minimum: This might be an edge case but if you ask `exists(ScanJob, identifier=42, name='Jar Jar Binks')` it may return `True` thanks to finding a `ScanJob` with the matching identifier even if the job has a different name.

tests/unit/io/test_scanning_store.py
```diff
@@ -46,63 +45,47 @@
...
         scanner = Scanner("Scanner two", "2")
         with pytest.raises(DuplicateNameError):
-            scanning_store.add_scanner(scanner)
+            scanning_store.add(scanner)
 
     def test_get_scanner(self, scanning_store):
-        assert scanning_store.get_scanner(
-            '9a8486a6f9cb11e7ac660050b68338ac') == SCANNER_ONE
+        assert scanning_store.get(
+            Scanner, '9a8486a6f9cb11e7ac660050b68338ac') == SCANNER_ONE
```
local-minimum: Perhaps use `SCANNER_ONE.identifier` as in the other test?

scanomatic/io/scanning_store.py
gdetrez: I don't like `_`

scanomatic/io/scanning_store.py
gdetrez: Good point

gdetrez:
> The exists method has identifier as a keyword argument but when used it is always supplied as a positional argument. I don't know if such things are of any major concern but each time I cause them myself it kind of scratches a little "may I shouldn't be doing this" kind of feeling.

This allow to call `exists` in a similar way as both `get` and `find`.

gdetrez:
> Second thing isn't really much of a problem now, maybe none at all, but the way add method both check general validity of adding and has specific checks for different types. The latter feels somewhat misplaced especially if the number of checks started to grow. But I don't have a clear idea of how to do it differently either...

This is not ideal right now. The existing checks will ultimately be done by the db engine. Optionally, they could also be done in the domain logic too, e.g. if we want better error messages.

scanomatic/io/scanning_store.py
local-minimum: agreed, which is why I've been writing out `identifier` as a an `id`-solution.

local-minimum:
> > The exists method has identifier as a keyword argument but when used it is always supplied as a positional argument. I don't know if such things are of any major concern but each time I cause them myself it kind of scratches a little "may I shouldn't be doing this" kind of feeling.
> 
> This allow to call exists in a similar way as both get and find.

Not the `find` method, and if the find-method actually did do do it the `exists` could be reduced to

```python
for _ in self.find(klass, id_, **constraints):
    return True
return False
```

PR #294, opened 2018-02-21
Qc autoload project

local-minimum:
Introduces a link on the `ScanningJobPanel` to QC the project's analysis.
This makes the QC UI autoload it hiding the project selector.
If there is no analysis for the project yet, an error message is displayed.

Fixes a few things with Luciano's code because they were making it annoying debugging the current implementation.

Will not fix css lint issues as these lines shouldn't have been included in the PR to start, but editor made some trailing spaces and such changes when I thought it was making a simpler solution than the current solution.


tests/system/test_scanjob.py
```diff
@@ -24,6 +25,21 @@ def test_compile_with_given_directory(scanomatic, browser):
...
+    scanjobname = 'Scan job {}'.format(uuid.uuid4().hex)
+
+    scannerid = create_scanner(scanomatic)
+    scanjobid = create_scanjob(scanomatic, scannerid, scanjobname)
+
+    browser.get(scanomatic + '/experiment')
+    browser.find_element_by_id('job-' + scanjobid).find_element_by_link_text(
+        "QC project".format(scanjobname)
+    ).click()
+    sleep(2)
```
gdetrez: If possible, use [selenium waits](https://selenium-python.readthedocs.io/waits.html) instead of sleep in tests

PR #296, opened 2018-02-21
splitting docker services server side

dindinoh:


docker-compose.yml
```diff
@@ -1,15 +1,24 @@
 version: '2'
 
 services:
-  scanomatic:
+  scanomatic-ui:
     image: phenomique/scanomatic:latest
     privileged: true
```
local-minimum: needed?

docker-compose.yml
```diff
@@ -1,15 +1,24 @@
...
     ports: ["5000"]
     volumes:
       - scan-o-matic-home:/root/.scan-o-matic
       - somprojects:/somprojects
-      - /dev/bus/usb:/dev/bus/usb
+    command: scan-o-matic --no-browser
+      
+  scanomatic-server:
+    image: phenomique/scanomatic:latest
+    privileged: true
```
local-minimum: needed?

scanomatic/server/interface_builder.py
```diff
@@ -136,29 +136,6 @@ def _remove_som_server(wait_for_jobs_to_stop):
 
         return successful_stop
 
-    @_verify_admin
-    def _server_shutdown(self, user_id, wait_for_jobs_to_stop=False):
```
local-minimum: Check if the functions (`self._remove*`) used by these two removed are used by anyone else. My guess is not, and then they too should be removed.

scanomatic/server/interface_builder.py
```diff
@@ -136,29 +136,6 @@ def _remove_som_server(wait_for_jobs_to_stop):
...
-            self.logger.error("Unknown error shutting down Scan-o-Matic server")
-
-        return sanitize_communication(val)
-
-    @_verify_admin
-    def _server_restart(self, user_id, wait_for_jobs_to_stop=False):
-
-        self._remove_rpc_server()
-
-        self._restart_server_thread(wait_for_jobs_to_stop)
```
local-minimum: This too is probably not used now

docker-compose.yml
dindinoh: nope, but it is nice

docker-compose.yml
dindinoh: will remove, thanks.

scanomatic/io/rpc_client.py
```diff
@@ -16,6 +16,7 @@
 # METHODS
 #
 
+socket.setdefaulttimeout(3)
```
gdetrez: Why is this needed?
At the very least, I would like a comment explaining why it is there but if the goal is to give the rpc server container time to boot, retrying might be a better solution.

scanomatic/io/app_config.py
```diff
@@ -174,10 +174,8 @@ def apply_local_rpc_settings(self):
                 "Could not read from '{0}',".format(self._paths.config_rpc) +
                 "though local settings were indicated to exist")
 
-        self._settings.rpc_server.host = Config._safe_get(
-            rpc_conf, "Communication", "host", '127.0.0.1', str)
-        self._settings.rpc_server.port = Config._safe_get(
-            rpc_conf, "Communication", "port", 12451, int)
+        self._settings.rpc_server.host = 'scanomatic-backend'
+        self._settings.rpc_server.port = 12451
```
gdetrez: Is there a way to overwrite this?
On tokaimura we have two Scan-o-matic instance running in the same compose file, how is it going to work there?

scanomatic/io/rpc_client.py
dindinoh: Not needed now, Will be removed.

PR #302, opened 2018-02-26
Simplify setup.py

gdetrez:
- remove everything but `setup()`
- use setuptools' `find_packages()` to automatically build the list of packages.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,123 +2,84 @@
 from __future__ import absolute_import
 
 import os
-import sys
-from subprocess import Popen, PIPE, call
-import json
-from logging import getLogger
 
-package_dependencies = [
-    'alembic',
-    'chardet',
-    'enum34',
-    'flask',
-    'flask-restful',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -2,123 +2,84 @@
...
-            'License :: OSI Approved :: GNU General Public License v3 (GPLv3)',
-            'Natural Language :: English',
-            'Operating System :: POSIX :: Linux',
-            'Programming Language :: Python :: 2.7',
-            'Topic :: Scientific/Engineering :: Bio-Informatics'
+    The package contains a user interface as well as an extensive package
+    for yeast colony analysis from scanned images.
+    """,
+    author="Martin Zackrisson",
+    author_email="martin.zackrisson@gu.se",
```
skymandr: I'm pretty sure this bounces...

gdetrez: I think I'll let @local-minimum update this if and how he wants to.

PR #303, opened 2018-02-26
Add scans and scan jobs to the database

gdetrez:
- Add a new tables `scans` and `scanjobs`
- Add a `ScanJobStore` class
- Add a `ScanStore` class
- Update API

This also changes the way the database schema is defined: instead of defining the schema twice, once in the migration scripts and once in `tables.py`, use sqlalchemy reflection abilities to get the schema from the database itself.  Now we can remove `tables.py`.


scanomatic/data/migrations/versions/39da65401ca3_create_scanjob_table.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+from __future__ import absolute_import
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects.postgresql import ExcludeConstraint
+from sqlalchemy.sql import quoted_name
+
+
+# revision identifiers, used by Alembic.
+revision = '39da65401ca3'
```
skymandr: when do these need updating? is it automatic or do we need to remember to do it?

gdetrez: The migration script should almost never change. If you need to change the schema, you create a new one using `alembic revision`.

scanomatic/data/scanstore.py
```diff
@@ -0,0 +1,49 @@
...
+        else:
+            raise KeyError(id_)
+
+    def get_all_scans(self):
+        query = self._table.select()
+        return self._get_scans(query)
+
+    def _get_scans(self, query):
+        for row in self._connection.execute(query):
+            yield Scan(
```
local-minimum: Just a thought, but not really suggesting you'd do it, but what would you think of having this be something like `yield Scan.from_dict(row)` and put the logic on that class instead. Same from the others (`ScanJob`).

scanomatic/data/migrations/versions/39da65401ca3_create_scanjob_table.py
```diff
@@ -0,0 +1,59 @@
...
+    op.create_table(
+        'scanjobs',
+        sa.Column('id', sa.Text(), primary_key=True),
+        sa.Column('name', sa.Text(), unique=True, nullable=False),
+        sa.Column('duration', sa.Interval(), nullable=False),
+        sa.Column('interval', sa.Interval(), nullable=False),
+        sa.Column(
+            'scanner_id',
+            sa.Text(),
+            sa.ForeignKey('scanners.id', name='fk_scanjob_scanner_id'),
```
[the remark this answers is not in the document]
gdetrez: "foreign key"

scanomatic/data/migrations/versions/da56e259f2e6_create_scans_table.py
```diff
@@ -0,0 +1,37 @@
...
+# revision identifiers, used by Alembic.
+revision = 'da56e259f2e6'
+down_revision = '39da65401ca3'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        'scans',
```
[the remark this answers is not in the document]
gdetrez: Well, as it is now, we can't put the image in the database because it would break the analysis pipeline.
We might put them there in the future but I'm not sure we'll want to. I'm not sure what would be the benefits.

scanomatic/data/scanstore.py
gdetrez: I don't think that's a good design as it means the domain data structures have to know about the database schema.

scanomatic/data/scanjobstore.py
```diff
@@ -0,0 +1,83 @@
...
+from sqlalchemy.sql.expression import exists
+
+from scanomatic.models.scanjob import ScanJob
+
+
+class ScanJobStore(object):
+    class IntegrityError(Exception):
+        pass
+
+    def __init__(self, connection, metadata):
```
local-minimum: This one and its parallels.

gdetrez: ok

PR #308, opened 2018-02-28
Store the scanner status in the database

gdetrez:
This is the last part of moving the scanning_store to a persistent database: migrating the scanner status.

This is a bit more aggressive refactoring: since the only thing we actually need from the status in the UI is the last time a status update was received, we only store this value in the `scanners` table (in a new column called `last_seen`). The actual values in the status update are not kept (some of them are still used to update metrics though.)

As a consequence, the `GET /api/scanners/ID/status` endpoint is removed (but it wasn't used).


scanomatic/models/scanner.py
```diff
@@ -2,7 +2,12 @@
...
-Scanner = namedtuple(
-    'Scanner',
-    ['name', 'identifier']
+ScannerBase = namedtuple(
+    'ScannerBase',
+    ['name', 'identifier', 'last_seen']
 )
+
+
+class Scanner(ScannerBase):
```
local-minimum: This is only to have the possibility of omitting `last_seen` and have it default to `None`? Is that ever needed?

scanomatic/scanning/update_scanner_status.py
```diff
@@ -92,7 +83,7 @@ def _add_scanner(scannerstore, scanner_id):
     try:
         name = get_generic_name()
         scannerstore.add(Scanner(name, scanner_id))
```
local-minimum: Should this not have a `now`-value for last-seen to given your other changes?

scanomatic/ui_server/scanners_api.py
```diff
@@ -18,49 +18,37 @@
...
-            datetime.now(pytz.utc)
-            - scanning_store.get_latest_scanner_status(
-                scanner_id).server_time
-            < SCANNER_TIMEOUT
-        )
-    except AttributeError:
-        return False
+def _scanner_is_online(scanner):
+    return (
+        scanner.last_seen is not None
```
local-minimum: See my other comment about if there is any reason why `last_seen` should be `None`-able.

scanomatic/models/scanner.py
gdetrez: Yes and no. We don't have a use for it now but some of the tests assume that it is possible to have a scanner that has not sent a status yet, so I leave it so.

PR #311, opened 2018-03-05
Remove `__CCC` global

gdetrez:
Refactor `scanomatic.data_processing.calibration` to remove the global `__CCC` global.

This doesn't really change the way it works: calibrations are still stored in an in-memory dictionary that is saved to disk explicitly in some of the functions. But instead of being a global dictionary, the calibration store is now encapsulated in a custom class (with dictionary-ish methods), passed to the calibration functions explicitly and referenced from the flask app.

A "store" is also created when needed in the different analysis jobs that used the `calibration` module.  If I understand thing correctly, this has the side benefits that it is no longer necessary to restart the backend when adding a new calibration.

### DIFF
--- a/scanomatic/data_processing/calibration.py
+++ b/scanomatic/data_processing/calibration.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+
 from collections import namedtuple
 from enum import Enum
 from itertools import izip
--- a/scanomatic/data_processing/calibration.py
+++ b/scanomatic/data_processing/calibration.py
@@ -1,28 +1,27 @@
-import numpy as np
+from collections import namedtuple
 from enum import Enum
 from itertools import izip
-from scipy.optimize import leastsq
-from scipy.stats import linregress
 import re
 from uuid import uuid1
-from collections import namedtuple
 
+import numpy as np
+from scipy.optimize import leastsq
+from scipy.stats import linregress
 
 from scanomatic.generics.maths import mid50_mean
+from scanomatic.image_analysis.first_pass_image import FixtureImage
+from scanomatic.image_analysis.image_basics import (
+    Image_Transpose, load_image_to_numpy
+)
+from scanomatic.io.ccc_data import (
+    CalibrationEntryStatus, CCCImage, CCCMeasurement, CCCPlate, CCCPolynomial,
+    CellCountCalibration, get_empty_ccc_entry, get_empty_image_entry,
+    get_polynomal_entry, load_cccs, save_ccc, validate_polynomial_format
+)
+from scanomatic.io.fixtures import Fixtures
 from scanomatic.io.logger import Logger
 from scanomatic.io.paths import Paths
-from scanomatic.io.fixtures import Fixtures
-from scanomatic.image_analysis.image_basics import (
-    load_image_to_numpy, Image_Transpose)
-from scanomatic.image_analysis.first_pass_image import FixtureImage
-from scanomatic.io.ccc_data import (
-    CCCImage, CCCPlate, CCCPolynomial, CCCMeasurement, CellCountCalibration,
-    CalibrationEntryStatus, get_empty_ccc_entry, load_cccs,
-    get_polynomal_entry, get_empty_image_entry, save_ccc,
-    validate_polynomial_format,
-)
 
-__CCC = {}
 _logger = Logger("CCC")
 
 CalibrationData = namedtuple(
@@ -52,11 +51,11 @@ class CalibrationValidation(Enum):
     BadData = 4
 
 
-def _ccc_edit_validator(identifier, **kwargs):
+def _ccc_edit_validator(store, identifier, **kwargs):
 
-    if identifier in __CCC:
+    if identifier in store:
 
-        ccc = __CCC[identifier]
+        ccc = store[identifier]
 
         if ("access_token" in kwargs and
                 ccc[CellCountCalibration.edit_access_token] ==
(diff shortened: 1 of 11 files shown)


scanomatic/models/factories/scanning_factory.py
```diff
@@ -420,7 +424,7 @@ def _validate_cell_count_calibration_id(cls, model):
 
         :type model: scanomatic.models.scanning_model.ScanningModel
         """
-        if model.cell_count_calibration_id in get_active_cccs():
+        if model.cell_count_calibration_id in get_active_cccs(CalibrationStore()):
```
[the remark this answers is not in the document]
gdetrez: fixed

scanomatic/data_processing/calibration.py
```diff
@@ -1,28 +1,27 @@
-import numpy as np
+from collections import namedtuple
```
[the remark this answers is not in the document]
gdetrez: fixed

scanomatic/data_processing/calibration.py
```diff
@@ -733,5 +715,37 @@ def get_all_colony_data(identifier):
...
+            if ccc[CellCountCalibration.identifier] in self._CCC:
+                _logger.error(
+                    "Duplicated identifier {0} is not allowed!".format(
+                        ccc[CellCountCalibration.identifier])
+                )
+            else:
+                self._CCC[ccc[CellCountCalibration.identifier]] = ccc
+
+    def keys(self):
+        return self._CCC.keys()
```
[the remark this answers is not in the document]
gdetrez: wontfix

scanomatic/data_processing/calibration.py
```diff
@@ -733,5 +715,37 @@ def get_all_colony_data(identifier):
...
+
+    def itervalues(self):
+        return self._CCC.itervalues()
```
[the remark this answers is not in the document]
gdetrez: wontfix

scanomatic/data_processing/calibration.py
```diff
@@ -733,5 +715,37 @@ def get_all_colony_data(identifier):
...
+        return iter(self._CCC)
+
+    def __getitem__(self, key):
+        return self._CCC[key]
+
+    def __setitem__(self, key, value):
+        self._CCC[key] = value
+
+    def iteritems(self):
+        return self._CCC.iteritems()
```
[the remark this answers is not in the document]
gdetrez: wontfix

scanomatic/io/ccc_data.py
```diff
@@ -67,13 +67,13 @@
...
-from types import StringTypes
 from enum import Enum
-from uuid import uuid1
+from glob import iglob
+import json
 import os
 import re
-import json
-from glob import iglob
+from types import StringTypes
```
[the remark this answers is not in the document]
gdetrez: wontfix

scanomatic/models/factories/analysis_factories.py
```diff
@@ -1,10 +1,13 @@
 import os
-from types import StringTypes, ListType, DictType
+from types import DictType, ListType, StringTypes
```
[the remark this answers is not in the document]
gdetrez: wontfix

tests/unit/data_processing/test_calibration.py
```diff
@@ -502,58 +502,54 @@ def test_delete_editable_ccc(self, edit_ccc):
...
-        calibration.add_ccc(ccc2)
-
-    def teardown_method(self):
-        calibration.reload_cccs()
+    @pytest.fixture(scope='function', autouse=True)
+    def setup(self, store):
+        with mock.patch(
+            'scanomatic.data_processing.calibration.save_ccc_to_disk',
+            return_value=True
+        ) as my_mock:
```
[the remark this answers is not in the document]
gdetrez: fixed

scanomatic/ui_server/data_api.py
```diff
@@ -803,7 +803,8 @@ def image_detect_colony():
 
         gc = GridCell(
             identifier,
-            get_polynomial_coefficients_from_ccc('default'),
+            get_polynomial_coefficients_from_ccc(
+                CalibrationStore(), 'default'),
```
local-minimum: Why does some api use `getcalibrationstore` and some use `CalibrationStore()`?

scanomatic/ui_server/calibration_api.py
```diff
@@ -3,26 +3,31 @@
...
+from scanomatic.models.factories.analysis_factories import AnalysisModelFactory
 from .general import (
-    serve_numpy_as_image, get_grayscale_is_valid, valid_array_dimensions,
-    json_abort
+    get_grayscale_is_valid, json_abort, serve_numpy_as_image,
+    valid_array_dimensions
 )
 
 
+def getcalibrationstore():
```
local-minimum: please separate words with `_`.

scanomatic/data_processing/calibration.py
```diff
@@ -52,11 +53,11 @@ class CalibrationValidation(Enum):
     BadData = 4
 
 
-def _ccc_edit_validator(identifier, **kwargs):
+def _ccc_edit_validator(store, identifier, **kwargs):
```
local-minimum: I'm wondering if you don't think this is much of a benefit to have then perhaps  remove this and `_validate_ccc_edit_request` while at it? Or if not while at it, as separate PR.

scanomatic/data_processing/calibration.py
```diff
@@ -733,5 +717,37 @@ def get_all_colony_data(identifier):
     }
 
 
-if not __CCC:
-    reload_cccs()
+class CalibrationStore(object):
```
local-minimum: I'm slightly confused, if instantiated and stored on the app (as it is further down, i think?) then it will diverge from the state on the disk. Isn't that a problem?

Why not instead scrap the `self._CCC` entirely have `_reload_cccs` be `_get_cccs` and return the dict. Then all public functions of the class just operate directly on that returned stuff from `_get_cccs`? It will be slower of course, but perhaps OK?

scanomatic/ui_server/data_api.py
gdetrez: Since this API only uses the default ccc, it doesn't really matter which it uses. But it's probably better that it uses the one on the app. I'll fix.

scanomatic/ui_server/calibration_api.py
gdetrez: This is consistent with how I called the other `get*store` methods.

scanomatic/data_processing/calibration.py
gdetrez: I thought about it but it implies many small refactorings. I think it's out of scope of the current task.

scanomatic/data_processing/calibration.py
gdetrez: I might not have been clear in the description. The PR is only refactoring the way the calibration store is accessed, it doesn't aim at solving the concurrency problem yet.

scanomatic/ui_server/calibration_api.py
local-minimum: alright, but I'd prefer that we avoid it in the future and consider refactoring since it lowers readability

PR #312, opened 2018-03-06
Import absolute_import everywhere

gdetrez:


gdetrez:
Note: I'm planning to ignore everything stickler says in this PR :smiling_imp:

gdetrez:
25tvcq

local-minimum:
You have to resolve the conflicts, but otherwise I'm thinking that it should be OK. Have you tried starting SoM just to see that nothing obviously bad happens?

gdetrez:
Conflicts fixed.
I'm not sure what "obviously bad" means... I checked that the system starts, and different pages are up but I haven't run more tests than that.

local-minimum:
If you really wanna go the full stretch, check that an experiment runs from compilation to feature extraction

Issue #147, opened 2017-08-25
CCC api may have concurrecy issues


skymandr:
From #146:

> Finally, and that's a big one, I think that there are concurrency issues here. It's not only the delete function, all the endpoint that manipulate the CCCs are likely to be affected. But it includes the delete one.
> A relatively likely scenario is that the user double clicks on the delete button (because too much :coffee:? ;-)) which triggers two runs of delete_non_deployed_calibration.
> One problem is that, if the stars are right, the parallel executions can be interleaved in a way that results in the CCC in memory being different than the one on disk. The other issue is that running save_ccc_to_disk in parallel will likely ends up in a garbled, unreadable file.

DoD: Make sure the endpoints are safe from over caffeinated users.

skymandr:
Would appreciate comments by @local-minimum and @gdetrez if you have any.

skymandr:
Proposed solution after IRL discussion: start using databases instead of filesystem for configuration/settings etc. Postgresql is a promising candidate. Currently, however, the service is running in single thread mode, so this is not an issue until it is scaled up for truly networked use.

gdetrez:
This should now be *mostly* solved. I suggest we open new issues in case we find problem with it.

PR #313, opened 2018-03-06
Start fleshing out the CalibrationStore interface

gdetrez:
Some more progress towards #147:
- remove the dictionary interface of `CalibrationStore` and replace it with more explicit methods
- start moving the responsibility of saving the CCCs to `CalibrationStore`. Only simple ones for now but I suspect that I'll want/need to do larger refactorings for the other ones so they can be separate PRs.

This is a refactoring, it is not intended to change the behavior at this point.

### DIFF
--- a/scanomatic/data_processing/calibration.py
+++ b/scanomatic/data_processing/calibration.py
@@ -54,30 +54,21 @@ class CalibrationValidation(Enum):
 
 
 def _ccc_edit_validator(store, identifier, **kwargs):
-
-    if identifier in store:
-
-        ccc = store[identifier]
-
+    if store.has_calibration_with_id(identifier):
+        ccc = store.get_calibration_by_id(identifier)
         if ("access_token" in kwargs and
                 ccc[CellCountCalibration.edit_access_token] ==
                 kwargs["access_token"] and
                 kwargs["access_token"]):
-
             if (ccc[CellCountCalibration.status] ==
                     CalibrationEntryStatus.UnderConstruction):
-
                 return True
-
             else:
-
                 _logger.error(
                     "Can not modify {0} since not under construction".format(
                         identifier)
                 )
-
         else:
-
             _logger.error(
                 "Bad access token for {}, request refused using {}"
                 .format(
@@ -86,7 +77,6 @@ def _ccc_edit_validator(store, identifier, **kwargs):
                 )
             )
     else:
-
         _logger.error("Unknown CCC {0}".format(identifier))
     return False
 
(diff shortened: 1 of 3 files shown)


scanomatic/data_processing/calibration.py
```diff
@@ -119,50 +109,31 @@ def _get_ccc_identifier(store, species, reference):
     if not species or not reference:
         return None
 
-    if any(True for ccc in store.itervalues() if
+    if any(True for ccc in store.get_all_calibrations() if
```
local-minimum: I don't really like how this whole function works, it's a bit obscure and arbitrary. Would feel better to have a function for checking if the ccc is valid (has both species and reference and there is no other ccc with exactly the same species and reference combo). And next just have the id be a uuid or key in a table that enumerates. But perhaps that is next PR?

scanomatic/data_processing/calibration.py
```diff
@@ -172,23 +143,21 @@ def get_polynomial_coefficients_from_ccc(store, identifier):
...
 def add_ccc(store, ccc):
-
-    if (ccc[CellCountCalibration.identifier] and
-            ccc[CellCountCalibration.identifier] not in store):
-
-        store[ccc[CellCountCalibration.identifier]] = ccc
-
-        save_ccc_to_disk(store, ccc)
+    if (
+        ccc[CellCountCalibration.identifier] and
```
local-minimum: In the vein of my previous comment, it feels like it is here that the ccc should get its identifier rather than checking that it is valid.

scanomatic/data_processing/calibration.py
```diff
@@ -210,47 +179,37 @@ def has_valid_polynomial(ccc):
...
 def delete_ccc(store, identifier):
-
-    ccc = store[identifier]
-
-    ccc[CellCountCalibration.status] = CalibrationEntryStatus.Deleted
-    ccc[CellCountCalibration.edit_access_token] = uuid1().hex
-
-    return save_ccc_to_disk(store, ccc)
+    ccc = store.get_calibration_by_id(identifier)
+    store.set_calibration_status(identifier, CalibrationEntryStatus.Deleted)
```
local-minimum: Who is here responsible to ensuring that the ccc hasn't been activated yet? Maybe that's the decorator, but maybe the decorator should have a better name to reflect that. And anyways I thought you wanted to actually remove the cccs that were under construction and got deleted instead of flagging them? In that and perhaps any case, maybe it should be `store.delete(identifier)`?

scanomatic/data_processing/calibration.py
```diff
@@ -210,47 +179,37 @@ def has_valid_polynomial(ccc):
...
-
+    if store.has_calibration_with_id(ccc[CellCountCalibration.identifier]):
         return save_ccc(ccc)
-
     else:
-
-        _logger.error("Unknown CCC identifier {0} ({1})".format(
-            ccc[CellCountCalibration.identifier], store.keys()))
+        _logger.error(
+            "Unknown CCC identifier %s",
```
local-minimum: I'm no fan of the `%` formatting... and since SoM generally uses `{}` I'd prefer consistency.

scanomatic/data_processing/calibration.py
gdetrez: As I see it, that's not really related to the current task.

scanomatic/data_processing/calibration.py
gdetrez: Maybe later, I haven't decided yet.

scanomatic/data_processing/calibration.py
gdetrez: I share your preference for the new format but logging is a bit of a special case as the `%` syntax is handled by the logging library in a smarter way.

local-minimum: Alright that's a valid exception. Pity it doesn't support both.

scanomatic/data_processing/calibration.py
local-minimum: I checked and the decorator ensures the status of the ccc.

PR #314, opened 2018-03-07
Add one more method to CalibrationStore and proper unit tests for the refactored use-cases

gdetrez:
- Add a method to `CalibrationStore` to get all the measurements for a CCC.
- Add unit tests to the refactored use cases (`add_ccc`, `activate`ccc`, `delete_ccc` and `construct_polynomial`) that mock the calibration store.

### DIFF
--- a/scanomatic/data_processing/calibration.py
+++ b/scanomatic/data_processing/calibration.py
@@ -530,20 +530,20 @@ def get_calibration_optimization_function(degree=5):
 
     coeffs = np.zeros((degree + 1,), np.float)
 
-    def poly(data_store, *guess):
+    def poly(measurements, *guess):
         coeffs[:-1] = np.exp(guess)
         return tuple(
             (np.polyval(coeffs, values) * counts).sum()
             for values, counts in
-            zip(data_store.source_values, data_store.source_value_counts))
+            zip(measurements.source_values, measurements.source_value_counts))
 
     return poly
 
 
 def get_calibration_polynomial_residuals(
-        guess, colony_sum_function, data_store):
+        guess, colony_sum_function, measurements):
 
-    return data_store.target_value - colony_sum_function(data_store, *guess)
+    return measurements.target_value - colony_sum_function(measurements, *guess)
 
 
 def get_calibration_polynomial(coefficients_array):
@@ -561,7 +561,7 @@ def poly_as_text(poly):
     return "y = {0}".format(" + ".join(coeffs()))
 
 
-def calculate_polynomial(data_store, degree=5):
+def calculate_polynomial(measurements, degree=5):
 
     fit_function = get_calibration_optimization_function(degree)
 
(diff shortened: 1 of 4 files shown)


scanomatic/data_processing/calibration.py
```diff
@@ -495,21 +495,15 @@ def validate_polynomial(slope, p_value, stderr):
     return CalibrationValidation.OK
 
 
-def _collect_all_included_data(ccc):
+def _collect_all_included_data(store, calibrationid):
```
local-minimum: why not stick with `identifier` as in the other functions or at least just have one name for the same identifier in the module.

scanomatic/data_processing/calibration.py
```diff
@@ -604,8 +598,7 @@ def calculate_polynomial(data_store, degree=5):
 @_validate_ccc_edit_request
 def construct_polynomial(store, identifier, power):
 
-    ccc = store.get_calibration_by_id(identifier)
-    data_store = _collect_all_included_data(ccc)
+    data_store = _collect_all_included_data(store, identifier)
```
local-minimum: I'm wondering if it becomes a bit more confusing now that there's a `store` and a `data_store` (which is more a collection of some data) ...

tests/unit/data_processing/test_calibration.py
```diff
@@ -932,3 +939,239 @@ def test_doesnt_scramble_data_while_sorting(self, store, edit_ccc):
...
+        ccc[CellCountCalibration.polynomial] = (
+            ccc_data.get_polynomal_entry(len(polynomial) - 1, polynomial)
+        )
+    ccc[CellCountCalibration.edit_access_token] = access_token
+    if active:
+        ccc[CellCountCalibration.status] = CalibrationEntryStatus.Active
+    else:
+        ccc[CellCountCalibration.status] = (
+            CalibrationEntryStatus.UnderConstruction
+            )
```
local-minimum: indentation errror?

tests/unit/data_processing/test_calibration.py
```diff
@@ -932,3 +939,239 @@ def test_doesnt_scramble_data_while_sorting(self, store, edit_ccc):
...
+        ccc = make_calibration(identifier=None)
+        assert calibration.add_ccc(store, ccc) is False
+        store.add_calibration.assert_not_called()
+
+    def test_duplicate_id(self):
+        store = MagicMock(calibration.CalibrationStore)
+        store.has_calibration_with_id.return_value = True
+        ccc = make_calibration()
+        calibration.add_ccc(store, ccc)
+        assert store.add_calibration.called_with(ccc)
```
local-minimum: What is it that happens here? The newer overwrites the older?

tests/unit/data_processing/test_calibration.py
gdetrez: No. Might be a code style but I'm not sure we have a style guide documented anywhere for scan-o-matic...

tests/unit/data_processing/test_calibration.py
gdetrez: No, just a bad assertion. Fixed.

PR #316, opened 2018-03-08
Calibrationstore interface extraction

gdetrez:
I think I have almost extracted a sufficient interface from `calibration.py` to put in a separate module that only deals with storing and retrieving data. The only thing remaining is adding measurements. The plan is then to implement this module using the database.
(Note: that's only the data that was in the `.ccc` file. Images and numpy arrays are still saved in files and not included in the `CalibrationStore` interface.)

I have manually tested creating and activating a "hanna" calibration with this.

The interface has become larger that I wanted to. I might be able to simplify it a by refactoring the way calibration works but that will require changing the API and the client code so I'll save it for later.
I could also split in stores for the different "types" (i.e. calibrations, images, plates and measurements). But that will be a lot of stores to pass around. Opinions welcome :smile: 

This is based on #314 and #315, so you might want to review only 20e440e.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ jobs:
         firefox: latest
       before_install:
         - sudo pip install tox
-        - wget -N http://chromedriver.storage.googleapis.com/2.30/chromedriver_linux64.zip -P ~/
+        - wget -N http://chromedriver.storage.googleapis.com/2.36/chromedriver_linux64.zip -P ~/
         - unzip ~/chromedriver_linux64.zip -d ~/
         - rm ~/chromedriver_linux64.zip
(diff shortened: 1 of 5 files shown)


gdetrez:
> I have a general question though and that is where the responsibility of the integrity of the store should be handled. Right now it is by that edit request decorator as far as I understand the code. I'm wondering if it shouldn't be the store that ensures that edits aren't made to activated cccs?

No, that's better handled in the domain logic.

> Second question is what is happening with the ccc_data module, is it going away while this is databasing?

Right now, I'm keeping the "datastructures" defined in `ccc_data`. Wether the rest should stay, in particular the code that read `.ccc` file depends if you want to continue support this file format in the future.

PR #322, opened 2018-03-13
Put calibrations in the database

gdetrez:
Sorry, this is very long but I couldn't really do better without introducing unused code.
I tried to bundle related changes in different commit which might make it easier to digest.

The main change here is re-implementing `CalibrationStore` using the database, that is:
- create the necessary tables in a new migration script
- add the new `CalibrationStore` in `scanomatic.data.calibrationstore`
- add tests

I also added the default calibration as part of the migration.

Then, in 92d58cc I remove the "old" `CalibrationStore` from `scanomatic.data_processing.calibration` and the corresponding unit tests.

In f543918, I refactor the calibration API to use the new store. That involved moving much of the corresponding tests to the integration suite.

Finally, integrate the new store with the different jobs that use calibrations (mostly to validate the passed ID...)
This was the most awkward part as I didn't find an easy way to pass the calibration store through all the factories and models to where it is needed without refactoring way too many things.  So I ended-up creating a context manager that creates a store from the environment variables directly. This is not ideal but not worst that the previous situation. And it makes it work without refactoring the whole rpc server. 

## Tests
I successfully went through the calibration process.
I successfully ran comile/analysis/feature extraction.


gdetrez:
Need to fix one more test...

PR #327, opened 2018-03-16
Scan job /terminate API

gdetrez:


tests/integration/api/conftest.py
```diff
@@ -59,6 +59,15 @@ def get_scan_job(self, jobid):
         def start_scan_job(self, jobid):
             return client.post('/scan-jobs/{}/start'.format(jobid))
 
+        def terminate_scan_job(self, jobid, message='Just testing'):
```
local-minimum: I don't think this really matters much but this was the api is only tested for terminating with a message yet from what I understood of the implementation, supplying a message is optional?

scanomatic/ui_server/serialization.py
```diff
@@ -20,6 +20,10 @@ def job2json(job):
     }
     if job.start_time is not None:
         obj['startTime'] = datetime2json(job.start_time)
+    if job.termination_time is not None:
+        obj['terminationTime'] = datetime2json(job.termination_time)
+    if job.termination_message is not None:
```
local-minimum: I'm wondering if this perhaps could/should be just `if job.termination_message` in the case that the UI supplies an empty string as the message when posting or is that better handled elsewhere?

PR #330, opened 2018-03-20
Delete scanjob API

gdetrez:
Add API endpoint `DELETE /api/scan-jobs/$ID` that removes the scanjob from the database. It checks that the scanjob hasn't started yet.

### DIFF
--- a/scanomatic/scanning/terminate_scanjob.py
+++ b/scanomatic/scanning/terminate_scanjob.py
@@ -2,12 +2,10 @@ from datetime import datetime
 
 from pytz import utc
 
+from .exceptions import UnknownScanjobError
+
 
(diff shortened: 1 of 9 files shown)


scanomatic/scanning/terminate_scanjob.py
```diff
@@ -2,12 +2,10 @@
...
 from pytz import utc
 
+from .exceptions import UnknownScanjobError
 
-class TerminateScanJobError(Exception):
-    pass
 
-
-class UnknownScanjobError(Exception):
+class TerminateScanJobError(Exception):
```
local-minimum: Very optional, but with the other file being called exceptions, it kind of feels like it would have all the exceptions of the scanning-submodule.

PR #343, opened 2018-04-11
Remove iframes

local-minimum:
* Makes SoM use templates instead.
* Removed the VERSIONTAG thingy, and if we want to invalidate cache by version in the future is suggest using `?ver={{ version }}`
* Removed some flashy animation since I figured it would be annoying if it happens whenever you navigated the page.
* Updated the name of one of the functions in `helper.js` since it collided with other function somewhere.
* Needed to add some hack to the init of the qc-view to hide the "talking to server"-thing.


local-minimum:
Don't want to fix stickler errors because they are mostly not due to my changes and those that are would have quite a few consequences if they are to be fixed.

local-minimum:
It was called `Map` before but that caused conflicts with some of our external js-libs on some of the views. I'm not sure it can be reverted.

gdetrez:
> It was called Map before but that caused conflicts with some of our external js-libs on some of the views. I'm not sure it can be reverted.

Oh, I see. I didn't see it in the other file, sorry. What about browser support for template literals?

PR #344, opened 2018-04-17
Pass docker password via stdin

gdetrez:
Attempting to fix the [current build failures on master](https://travis-ci.org/Scan-o-Matic/scanomatic/builds/366148818). Cannot really be tested without merging...


gdetrez:
I remove yamllint from stickler, it's just annoying.

PR #345, opened 2018-04-17
 Add "Projects" page scaffolding

gdetrez:
Add a new path `/projects` with a web page rendering a simple react component and a system test ensuring that everything works.

I also updated the babel config to enable object spread (i.e. `{ ...object  }`).


local-minimum:
LGTM, but no point accepting until tests pass because it will dismiss my review....

PR #351, opened 2018-04-19
Add storybook

gdetrez:
One common and annoying problem that arise when working on react component is that it is sometime necessary to go through many hoops just to see how a component looks like. E.g. to see a running scanning job in scan-o-matic, one needs an online scanner, and then to create a job with the correct scanner and start it.

One possible solution is to temporarily insert a component with the desired state in the page but that's not very practical or elegant.  So my idea wast to build a static page with a collection of components in different states but of course, npm already has a tool for that, which is [storybook](https://storybook.js.org/).

It's exactly that: it lets you build a collection of components in different states but it adds some widgets for navigation, auto-reload, a console... (and a gazillion dependencies, but not in production...).

 I think this should be helpful when developing/reviewing components but could maybe also replace many low-vlaue/high-maintenance tests like "&lt;MyComponent/&gt; should render a div with class foo and text bar". 

The current configuration auto-discovers stories from `components/*.stories.js` files.

My suggestion is that we try in in scan-o-matic for a few sprints. If we like it, we may then want to add it to other projects. If it turns out to be too much maintenance for too little value, having the stories in separate files should make it easier to remove the whole thing.  

### to try it

```
npm install
npm run storybook
```
And then http://localhost:9001

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,0 +1,13 @@
+import { configure } from '@storybook/react';
+
+const req = require.context(
+    '../scanomatic/ui_server_data/js/src/components',
+    true,
+    /\.stories\.jsx$/
+)
+
+function loadStories() {
+  req.keys().forEach((filename) => req(filename))
+}
+
+configure(loadStories, module);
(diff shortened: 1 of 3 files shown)


gdetrez:
Did you try the instructions above?

PR #354, opened 2018-04-23
Add bootstrap css in project page

gdetrez:


gdetrez:
> I was wondering if it is preferred to have them here in the html or if it is better to have them in the js as import.

That's a good question. I think for the bootstrap file, since it is included in many pages, having it separated means that the browser can cache it.

PR #356, opened 2018-04-23
Redux store for projects page

gdetrez:
Add action creators, reducers and selectors to manage the state of the projects page.  Right now, there's no communication with the server: the state only exists in redux and doesn't survive a page reload.

In addition to the standard actions, this also adds a "thunk".  Thunks are delayed actions that have the possibilities of doing side effects or dispatch multiple actions. Here the `submitNewProject` action is a thunk that checks the form values and dispatches action to mark the form as submitted, add project and close the form.

Action, reducers and selectors are also statically typechecked with [flow](http://flow.org/).  Most notably, because actions and reducers are typechecked, we don't have to import/export a gazillion  of action type constants but still get the guarantee that there are no typos/discrepancies between the action types in `actions.js` and the reducers. (We actually get even better guarantees!) This is a bit of an experiment but so far I mostly like it.


scanomatic/ui_server_data/js/src/projects/reducers/forms/newProject.js
```diff
@@ -0,0 +1,34 @@
...
+        name: '',
+        description: '',
+    },
+};
+
+export default function reducer(state: State = null, action: Action): State {
+    switch (action.type) {
+    case 'NEWPROJECT_INIT':
+        return defaultState;
+    case 'NEWPROJECT_CHANGE':
```
local-minimum: No need to type-check here?

scanomatic/ui_server_data/js/src/projects/reducers/forms/newProject.js
```diff
@@ -0,0 +1,34 @@
...
+        if (state == null) return null;
+        return {
+            ...state,
+            fields: {
+                ...state.fields,
+                [action.field]: action.value,
+            },
+        };
+    case 'NEWPROJECT_SUBMIT':
+        if (state == null) return null;
```
local-minimum: If my question above applied, it does so too here.

scanomatic/ui_server_data/js/src/projects/reducers/forms/index.js
```diff
@@ -0,0 +1,8 @@
+// @flow
+import { combineReducers } from 'redux';
+import newProject from './newProject';
+
+const forms = combineReducers({ newProject });
```
local-minimum: And I suppose `newExperiment` would go in here too?

scanomatic/ui_server_data/js/src/projects/selectors.js
```diff
@@ -0,0 +1,93 @@
+// @flow
+import type { State } from './state';
+
+type Project = {
+id: string,
```
local-minimum: not important: missing indent

scanomatic/ui_server_data/js/src/projects/actions.js
```diff
@@ -0,0 +1,120 @@
...
+    };
+}
+
+export function clearNewExperiment(): Action {
+    return {
+        type: 'NEWEXPERIMENT_CLEAR',
+    };
+}
+
+export function addExperiment(
```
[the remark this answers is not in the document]
gdetrez: For consistency I think `submitNewExperiment` should be the thunk action. But I think you got the idea.

scanomatic/ui_server_data/js/src/projects/reducers/entities/index.js
```diff
@@ -0,0 +1,6 @@
+// @flow
+import { combineReducers } from 'redux';
+import projects from './projects';
+
+const entities = combineReducers({ projects });
```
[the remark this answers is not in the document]
gdetrez: Right. It probably should, but I missed it.
Doesn't really matter until there are components to use it though.

scanomatic/ui_server_data/js/src/projects/reducers/forms/index.js
gdetrez: Correct :grinning:

scanomatic/ui_server_data/js/src/projects/reducers/forms/newExperiment.js
```diff
@@ -0,0 +1,61 @@
...
+                fields: {
+                    ...state.fields,
+                    [action.field]: action.value,
+                },
+            };
+        default:
+            return state;
+        }
+    case 'NEWEXPERIMENT_SUBMIT':
+        if (state == null) return state;
```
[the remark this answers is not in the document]
gdetrez: > Question: With the type-checking for e.g. interval above, I'm in a bit of responsibilities confusion. I'm thinking of if it's more proper to throw an error here since it is basically an invalid input to have a submit-action if the state is null or undefined.

Yes and no.  The way I see it is that an action with type `NEWEXPERIMENT_CHANGE` where `field=interval` and `value` is not a number is not a valid action.  On the other hand, `NEWEXPERIMENT_SUBMIT` is a valid action and it is my understanding that it should then be either ignored or handled by the reducer. But in any case, it shouldn't throw.

I think maybe I should refactor this: It'd probably be clearer if the type annotations captured the different field types and the dynamic checks where done in the action creator.

> Same thought hit me on NEWEXPERIMENT_INIT, if action.projectId were not in the state. But that seems more involved and messy.

None of the relationships in the state are enforced, which is unfortunate. I'm not sure the benefits would be worth the effort though.

scanomatic/ui_server_data/js/src/projects/reducers/forms/newProject.js
gdetrez: No, all fields take a string value.

PR #358, opened 2018-04-24
Duration input

local-minimum:
I tested that it worked with the scan-jobs ui manually too.

### DIFF
--- a/scanomatic/ui_server_data/js/src/components/DurationInput.jsx
+++ b/scanomatic/ui_server_data/js/src/components/DurationInput.jsx
@@ -8,29 +8,71 @@ export default class DurationInput extends React.Component {
         this.handleDaysChange = this.handleDaysChange.bind(this);
         this.handleHoursChange = this.handleHoursChange.bind(this);
         this.handleMinutesChange = this.handleMinutesChange.bind(this);
+        this.handleDaysFocus = () => this.setState({ daysFocus: true });
+        this.handleHoursFocus = () => this.setState({ hoursFocus: true });
+        this.handleMinutesFocus = () => this.setState({ minutesFocus: true });
+        this.handleDaysBlur = this.handleDaysBlur.bind(this);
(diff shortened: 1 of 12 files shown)


scanomatic/ui_server_data/js/src/components/DurationInput.jsx
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    handleNewDuration(duration) {
+        const { days, hours, minutes } = (duration || {});
+        const newDuration = Duration.fromDaysHoursMinutes(days || 0, hours || 0, minutes || 0);
+        const value = newDuration.totalMilliseconds;
+        if (value !== this.props.duration) {
+            if (this.props.onChange) this.props.onChange(value);
+        }
+        this.setState({
```
gdetrez: I think this breaks the one-way dataflow: the way you have defined your props indicate that you don't own the duration value but here you assume that you do. (for instance, I don't think the component will properly re-render if the duration prop changes.)

One way to fix it is that the values of days, hours and minutes should always come from props and only keep the temporary overwritten value for the current focused field in the state.

local-minimum: fixed

gdetrez:
i'm not completely sure this completely fixes the issue. If I recall correctly, using the little arrow on a number input produce a change event but no focus/blur. If that's correct, the overwritten value will be left in the state even though it probably shouldn't.
One way to fix it is to handle the focus event and handle a change differently depending on whether the field is focused or not.

Since this is non trivial, maybe add a few more complex scenario in the tests. E.g.
- focus, change, blur, update props, check values
- change, update props, check values
- focus, change an other field, update props, check values (should focused field change value?)
- focus, change, change an other field, update props, check values (same?)
- etc

local-minimum:
Added more complex tests. Perhaps not all you suggested because some I didn't fully grasp.

gdetrez:
Looks good enough for me. We can probably iron out corner cases later. :+1:

PR #360, opened 2018-04-24
Projects page store: add getScanners and submitExperiment

gdetrez:
Add `getScanner` selector and `submitNewExperiment` action.

### DIFF
--- a/scanomatic/ui_server_data/js/src/projects/selectors.js
+++ b/scanomatic/ui_server_data/js/src/projects/selectors.js
@@ -45,7 +45,7 @@ export function getProjects(state: State): Array<Project> {
 }
 
 export function getScanners(state: State): Array<{ name: string, id: string, isOnline: boolean, isFree: boolean }> {
-    const scanners = Object.keys(state.entities.scanners).sort().map(key => ({
+    const scanners = Object.keys(state.entities.scanners).map(key => ({
         ...state.entities.scanners[key],
         id: key,
     }));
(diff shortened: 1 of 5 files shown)


scanomatic/ui_server_data/js/src/projects/selectors.js
```diff
@@ -44,6 +44,19 @@ export function getProjects(state: State): Array<Project> {
     });
 }
 
+export function getScanners(state: State): Array<{ name: string, id: string, isOnline: boolean, isFree: boolean }> {
```
local-minimum: Maybe not important now, but I suppose the scanner should be a type?

scanomatic/ui_server_data/js/src/projects/selectors.js
```diff
@@ -44,6 +44,19 @@ export function getProjects(state: State): Array<Project> {
...
+    const scanners = Object.keys(state.entities.scanners).sort().map(key => ({
```
local-minimum: why sort if you resort below?

scanomatic/ui_server_data/js/src/projects/selectors.js
gdetrez: I don't think it should. It could but since types are not currently exported from this module, it really doesn't matter that much.

scanomatic/ui_server_data/js/src/projects/selectors.js
gdetrez: Good point, too much copy paste I guess...

PR #364, opened 2018-04-27
Feature extract dialogue

local-minimum:
screenshot from 2018-04-27 11-15-38
screenshot from 2018-04-27 11-15-50
screenshot from 2018-04-27 11-16-04

This is fully integrated with existing API and replaces the functionality of the `/feature_extract`

Button only visible on completed scan-jobs. Though this doesn't necessarily mean you can do feature extraction as there has to have been a compilation and an analysis too. I think this quirk is fine for now, for this iteration.


scanomatic/ui_server_data/js/src/components/ScanningJobPanel.jsx
```diff
@@ -36,13 +35,26 @@ class ScanningJobPanel extends React.Component {
...
 
-    handleConfirmStop(reason) {
+    handleShowFeatureExtractDialogue() {
+        this.setState({ dialogue: 'featureExtact' });
+    }
+
+    handleFeatureExtract(keepQC) {
         this.setState({ dialogue: null });
-        this.props.onStopJob(this.props.scanningJob.identifier, reason);
+        extractFeatures(this.props.scanningJob.identifier, 'analysis', keepQC)
```
gdetrez: I don't think this should be in a presentational component (importing `api` should have been a clue ;-))

local-minimum: Fair enough. It just felt like there was going to be so much passing stuff around. (The parameters and the callbacks).

local-minimum:
I think it is quite feasible to make system tests for this given some short cuts, specifically that we don't really want to check that a feature extraction gets started and is running. The latter would require an experiment directory and a full analysis directory in it. That fixture would need to be quite a few MB. To be able to run the system test though in any reasonable manner you have to prep the system with a completed experiment in the db and log in to the container and add an `analysis` folder ot that experiments' folder.

gdetrez:
I think for this feature, it's enough to check that you get the "... is enqueued" message in the UI (and maybe the corresponding message on errors).

tests/system/test_experiment_page.py
```diff
@@ -139,6 +145,30 @@ def stop_job(self, reason='', confirm=True):
...
+        )
+        if keep_qc:
+            self.element.find_element(*self.keep_qc_locator).click()
+        if confirm:
+            (
+                self.element
+                    .find_element(*self.extract_features_ok_locator)
+                    .click()
+            )
+            WebDriverWait(self.element, 2).until(
```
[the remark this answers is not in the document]
local-minimum: it is probably OK with 2 since it doesn't have to load a new page. But I'll update to 5.

PR #367, opened 2018-04-27
Refactor <NewExperimentPanel/>

gdetrez:
- Make `erros` a `Map` instead of an object
- Make the name of the scanner id more consistent with the prop name `scannerId`
  (`scanner -> scannerId` in errors and in the callback)
- Add a valid option in the scanner drop down for when `scannerId` is
  empty
- Leave sorting the scanners to the selector

### DIFF
--- a/scanomatic/ui_server_data/js/src/components/ProjectPanel.jsx
+++ b/scanomatic/ui_server_data/js/src/components/ProjectPanel.jsx
@@ -55,7 +55,7 @@ ProjectPanel.propTypes = {
 };
 
 ProjectPanel.defaultProps = {
-    newExperimentErrors: null,
+    newExperimentErrors: undefined,
     newExperiment: null,
     newExperimentActions: {},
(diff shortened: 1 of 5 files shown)


scanomatic/ui_server_data/js/src/components/ProjectPanel.jsx
```diff
@@ -55,7 +55,7 @@ ProjectPanel.propTypes = {
 };
 
 ProjectPanel.defaultProps = {
-    newExperimentErrors: null,
+    newExperimentErrors: undefined,
```
local-minimum: This change I don't understand.

gdetrez: Default props are only used on `undefined`, not on `null`.
This is to make `<NewExperimentPanel/>` use it's default `errors` prop, an empty map as default.

PR #368, opened 2018-04-27
Add dummy scanner reducer

gdetrez:
Add a dummy scanner reducer and add a scanner in the initial state. This is for testing the new experiment form.


gdetrez:
Well, it is but in a different PR. To many at the same time I guess...

PR #370, opened 2018-04-30
Run system test using sauce labs on travis

gdetrez:
This adds an extra stage to the travis build that runs the system tests using saucelabs browsers.
* it is configured to use two browser: Firefox and Chrome, both on windows (the available browsers on Linux are lagging behind quite a bit :disappointed:).
* It will only run on push to master.  There are a few reason for this: the credentials are not available on external pull requests ; I want to make sure it's not too flaky ; and more importantly, it takes 20 minutes to run :fearful:

### DIFF
--- a/tests/system/conftest.py
+++ b/tests/system/conftest.py
@@ -1,11 +1,12 @@
 from __future__ import absolute_import
 
-import warnings
+import os
 
 import py.path
 import pytest
 import requests
 from selenium import webdriver
+from selenium.common.exceptions import WebDriverException
 
 
 @pytest.fixture(scope='session')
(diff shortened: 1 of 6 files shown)


tests/system/conftest.py
```diff
@@ -48,14 +111,12 @@ def pytest_configure(config):
...
-        driver = webdriver.Chrome()
-    else:
-        raise ValueError('Unknown browser {}'.format(browser))
-    yield driver
-    driver.close()
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
+def pytest_runtest_makereport(item, call):
+    # this sets the result as a test attribute for SauceLabs reporting.
+    # execute all other hooks to obtain the report object
+    outcome = yield
```
local-minimum: I looked up pytest about this, but I just don't really get what is happening here at all.

gdetrez: This hooks into pytest report generation but we only want to access the report, not generate it. What the yield does is tell pytest to run the normal report generation and pass us the result.

You can look at it as a "fixture" but for the test report instead of the test itself.

The point here is to make the test outcome available in the `saucelabs` fixture (line 95).

PR #371, opened 2018-05-02
Project Panel now expandable

skymandr:
Project Panel now expandable


scanomatic/ui_server_data/js/src/components/ProjectPanel.jsx
```diff
@@ -3,43 +3,72 @@ import PropTypes from 'prop-types';
...
+        this.setState({ expanded: !this.state.expanded });
+    }
+
+    render() {
+        const {
+            name, description, onNewExperiment, newExperiment, newExperimentActions,
+            newExperimentErrors, scanners,
+        } = this.props;
+        let newExperimentPanel;
+        if (this.newExperiment) {
```
local-minimum: This should also check for if `expanded`.

PR #374, opened 2018-05-03
Link to analysis

local-minimum:
screenshot from 2018-05-03 10-51-38

Autofils relevant part of form just like the compile link.
Unit and system test.


tests/system/test_experiment_page.py
```diff
@@ -264,3 +295,21 @@ def test_requesting_feature_extract_on_job_without_analysis_causes_error(
...
+        reason = "I don't want to do this anymore!"
+        create_scanner(scanomatic, scannerid)
+        page = ExperimentPage(browser, scanomatic)
+        page.add_job(name=jobname, scannerid=scannerid)
+        panel = page.find_job_panel_by_name(jobname)
+        job_id = panel.job_id
+        panel.start_job()
+        panel.stop_job(confirm=True, reason=reason)
+        panel.click_analysis()
+        assert page.main_heading == 'Analysis'
```
gdetrez: This is not really the way the PageObject pattern is used.
Since you end-up in a different page, you should have a different PageObject class for it. And `click_analysis` should return an instance of it:
```
analysis_page = panel.click_analysis()
assert analysis_page.main_heading == 'Analysis'
```

tests/system/test_experiment_page.py
```diff
@@ -264,3 +295,21 @@ def test_requesting_feature_extract_on_job_without_analysis_causes_error(
...
+        form = AnalysisForm(page.driver)
```
gdetrez: Somewhat similar here, you should get `AnalysisForm` from the `AnalysisPage` class. Something like `analysis_form = AnalysisPage.get_analysis_form`.

In that case, since the page is pretty much just the form, it might not be so interesting to have a separate `AnalysisForm` class, and you could just have `AnalysisPage.compile_file`.
Also, try to use names that match the input labels. In this case, I'm not sure if `compile_file` should refer to "COMPILATION FILE" or "COMPILE INSTRUCTION FILE"

PR #377, opened 2018-05-07
Made ProjectPanel expansion depend on props

skymandr:


scanomatic/ui_server_data/js/src/components/ProjectPanel.jsx
```diff
@@ -19,7 +21,9 @@ class ProjectPanel extends React.Component {
             name,
             newExperimentDisabled,
             onNewExperiment,
+            defaultExpanded,
         } = this.props;
+        const expanded = this.state.expanded == null ? defaultExpanded : this.state.expanded;
```
local-minimum: Might be useful to have this as a small function since it is repeated and vital to the functionality that they are identical.

PR #378, opened 2018-05-07
Foldable experiments

skymandr:


skymandr:
Tests need repairing.

scanomatic/ui_server_data/js/src/components/ExperimentPanel/index.jsx
```diff
@@ -18,6 +18,16 @@ export default class ExperimentPanel extends React.Component {
...
         this.handleShowFeatureExtractDialogue = () => this.setState({ dialogue: 'extract' });
+        this.handleToggleExpand = this.handleToggleExpand.bind(this);
+    }
+
+    getExpanded() {
+        return this.state.expanded == null ? this.props.defaultExpanded : this.state.expanded;
+    }
+
+    handleToggleExpand() {
+        const expanded = this.getExpanded();
```
local-minimum: Since you have lint error on function-order you could just as well also make line 29 & 30 into a single line. A bit unnecessary verbose to assign a local const.

skymandr: More unnecessary than making two function calls?

skymandr: Never mind... You're right. I'll fix it in the other branch as well.

PR #381, opened 2018-05-08
Flow config

local-minimum:
This test-file causes error in flow. I wasn't sure if it was good idea to omit entire `node_modules` since it might be useful if we use modules that use flow too.


local-minimum:
@gdetrez  I will merge this right away hoping it is fine with you, if not we can fix it next week.

PR #382, opened 2018-05-08
Default expand firsts

local-minimum:
Most recent experiment in a project and most recent project are default expanded.


local-minimum:
The second travis had a random fail on that bad unit-test first and restarting that step triggered failing system-tests. I wonder if there's something we can do about this...

PR #384, opened 2018-05-08
Markdown support for descriptions

local-minimum:
screenshot from 2018-05-08 12-14-56

Doesn't support headings (probably good since it would break layout without some css considerations). Supports sub and super scripting.
Applies to both project and experiment descriptions.


local-minimum:
The solution needs setting dangerous html but markdown-it should be safe since the markdown module is not set to allow html-tags:

https://markdown-it.github.io/markdown-it/#MarkdownIt.new

local-minimum:
I did this to see what your thought were on it and you might be right that `<pre>` is sufficient for now.
I thought this markdown package did sanatize, but you are right we'd probably want to check it if we decided to do it... I'll close this PR and one of us make's the descriptions `<pre>` and make sure it looks decent.

PR #385, opened 2018-05-08
Pinning inputs and displays

local-minimum:
Only showing details of the form and the info-table.

# New experiment
screenshot from 2018-05-08 18-27-50
## Errors
screenshot from 2018-05-08 18-28-09

# Listing experiments
_This doesn't match the input above, here plate 3 is left empty, all other plates have stuff_
screenshot from 2018-05-08 18-27-28


scanomatic/ui_server_data/js/src/components/NewExperimentPanel.spec.jsx
```diff
@@ -26,6 +26,7 @@ describe('<NewExperimentPanel/>', () => {
                 identifier: 'haha',
             },
         ],
+        pinning: new Map([[1: null, 2: '384']]),
```
gdetrez: I forgot: I think this is not what you want (multiple instances).

local-minimum:
That should hopefully deal with all occurrences of `pinning`... they were a few.

gdetrez:
lgtm

PR #387, opened 2018-05-09
Displaying scanners status

local-minimum:
screenshot from 2018-05-09 14-06-31

This implements a component that would be a major part of or even entire status page (depending on how well analysis status etc of experiments will be showed on the projects page).

It only implements the component so before it is to become useful there's a need to connect it to the existing api and add a redux store.


local-minimum:
This is not for the projects-page but as first step in replacing the current status-page.

PR #388, opened 2018-05-16
Adjusted css to render line breaks in descriptions

skymandr:


skymandr:
Not sure how this should be tested, or if it should be, since it is mainly a CSS-thing...

local-minimum:
Include screenshots of it in action or a gif/movie

PR #390, opened 2018-05-23
New fixture editor

local-minimum:
screenshot from 2018-05-23 11-18-34
This re-implements part of the create fixtures UI without a redux-store. Since that is my next step, I would like to get feedback on these rather complex components (interactions-wise) getting ahead of myself with the store.


skymandr:
I think you should probably walk us through this one live.

local-minimum:
I'm just gonna abandon this PR for now. I'll keep the branch should we ever work on it again... or maybe I should just merge it into master because why not?

PR #394, opened 2018-05-31
Fixed left/right arrow buttons in QC

skymandr:
- Left/right buttons now behave as expected.
- Q-index is shown in the UI
- Added button to go to first Q-index

som_arrows


skymandr:
This PR grew a little in scope. Be back later.

PR #395, opened 2018-06-01
The store for QC

local-minimum:
This just creates the store for the QC and the actions/selectors needed for the integration.

I checked out the files from my integration-branch so I hope I got them all.


scanomatic/ui_server_data/js/src/qc/StateBuilder.js
```diff
@@ -0,0 +1,53 @@
+import type { State, TimeSeries, PlateOfTimeSeries, Plate, Settings } from './state';
```
gdetrez: You have types in this file but didn't activate flow.

scanomatic/ui_server_data/js/src/qc/StateBuilder.js
```diff
@@ -0,0 +1,53 @@
...
+        return this;
+    }
+
+    setFocus(plate: number, row: number, col: number) {
+        if (plate !== this.plate.number) return this;
+        this.plate.focus = { row, col };
+        return this;
+    }
+
+    setPlateGrothData(
```
skymandr: Growth

scanomatic/ui_server_data/js/src/qc/actions.js
```diff
@@ -0,0 +1,83 @@
...
+    return { type: 'PROJECT_SET', project };
+}
+
+export function setPinning(plate : number, rows: number, cols: number) : Action {
+    return {
+        type: 'PINNING_SET', plate, rows, cols,
+    };
+}
+
+export function setPlateGrothData(
```
skymandr: Growth

scanomatic/ui_server_data/js/src/qc/actions.js
```diff
@@ -0,0 +1,83 @@
...
+        if (plate == null) {
+            throw new Error('Cannot retrieve curves if project not set');
+        }
+
+        return getPlateGrowthData(project, plate).then((r) => {
+            const { smooth, raw, times } = r;
+            dispatch(setPlateGrothData(plate, times, raw, smooth));
+            const rows = raw.length;
+            const cols = raw[0].length;
+            dispatch(setPinning(plate, rows, cols));
```
gdetrez: My first reaction here is that this set pinnig should be in the reducer for `PLATE_GROWTHDATA_SET`. But it seems that the pinning only depends on the value of `raw`, which is in the state. So why not just make it a selector instead?

scanomatic/ui_server_data/js/src/qc/api.js
```diff
@@ -0,0 +1,16 @@
+// @flow
```
gdetrez: I'm not a big fan of this new file. It breaks the current organization and coupling the api client with the page doesn't make much sense to me.

If what you're after is to be able to enable flow for this function only I suggest that you:
- move `api.js` to `api/index.js`
- put this function in a file in `api/`. You can call it `getPlateGrothData.js` to avoid the dummy default export.
- re-export the function from `api/index.js` so that the rest of the code doesn't have to know about this rather arbitrary file split.

Since this new function doesn't really depend on the rest of the changes, maybe do that refactoring in a different PR :-)

scanomatic/ui_server_data/js/src/qc/reducers/plate.js
```diff
@@ -0,0 +1,29 @@
+import type { Action } from '../actions';
```
gdetrez: //@flow?

scanomatic/ui_server_data/js/src/qc/selectors.js
```diff
@@ -0,0 +1,44 @@
+// @flow
+
+import type { State, TimeSeries, Pinning, PlatePosition } from './state';
```
gdetrez: if you use those types as return type for the selector function, re-export them from this module . Otherwise this might introduce problematic dependencies.

local-minimum:
Integrated new API into the PR, ready for re-review.

scanomatic/ui_server_data/js/src/qc/StateBuilder.js
local-minimum: fixed

scanomatic/ui_server_data/js/src/qc/reducers/plate.js
local-minimum: fixed

PR #402, opened 2018-06-04
Qc integrate store

local-minimum:
Integrates the store for the curves with the UI.

Adds tests and hopefully resolves the rnd fail on build tests.


scanomatic/ui_server_data/js/qc_normDrawCurves.js
```diff
@@ -59,17 +57,16 @@ function DrawCurves(container, data, gt, gtWhen, yld) {
...
     //properties
-    var data;
     var margin;
     var height;
     var width;
     var generationTimeWhen;
     var generationTime;
     var growthYield;
-
+    let drawn = false;
```
skymandr: I think it's untidy mixing `let` and `var` like this. I see why you wouldn't want to change the `var`s in this PR, but would you perhaps consider cleaning this up in a separate PR?

local-minimum: I agree it's untidy, but out of scope to fix existing and I quite strongly do not agree with introducing `var`! If I need to add something new I should use `let` or `const`. Introducing `var` should be considered a no-no. Cleaining up `var` could be different PR.

scanomatic/ui_server_data/js/qc_normHelper.js
```diff
@@ -468,7 +469,7 @@ function renderPlate(phenotypePlates) {
     const path = phenotypePlates.url;
     const plateIdx = phenotypePlates.index;
     const project = $('#spProject').text();
-    console.log(`experiment: ${path}`);
+    console.log(`experiment: ${path}`, phenotypePlates);
```
skymandr: why we log so much :(

local-minimum: I know, but right now they also illustrate the issue with loading that some things are loaded as many as 3 times depending on conditions...which was why I dind't remove them where I found them.

PR #405, opened 2018-06-18
Integrate QC store quality index.

local-minimum:
This moves the control over what position/curve is focused/shown entirely into the redux store. In there it is entirely driven by the selected quality index.

I've mostly relied on the unit-tests of the store and bridge code and the system tests that existed. They cover the vast majority of functionality that is touched by this PR. There are some details like validating which position on the heatmap gets that circle mark at each time as well as the grid images thing. I opted to not extend them for this PR as it is rather huge. I can add them though, but it might be better to add them in their own PR in that case?

### DIFF
--- a/scanomatic/ui_server/templates/qc_norm.html
+++ b/scanomatic/ui_server/templates/qc_norm.html
@@ -80,7 +80,7 @@
                   if (this.checked) {
                       $("#divMarkStates").show();
                       $("#qidxHint").show();
-                      setExperimentByQidx(qIdxOperations.Current);
+                      showCurrentQIndex();
                   } else {
                       $("#divMarkStates").hide();
                       $("#qidxHint").hide();
@@ -151,13 +151,13 @@
                   markExperiment(plateMetaDataType.NoGrowth, true);
               });
               $("#btnQidxReset").click(function () {
-                  setExperimentByQidx(qIdxOperations.Reset);
+                  window.qc.actions.setQualityIndex(window.qc.selectors.getPlate(), 0);
               });
               $("#btnQidxNext").click(function () {
-                  setExperimentByQidx(qIdxOperations.Next);
+                  window.qc.actions.nextQualityIndex(window.qc.selectors.getPlate());
               });
               $("#btnQidxPrev").click(function () {
-                  setExperimentByQidx(qIdxOperations.Prev);
+                  window.qc.actions.previousQualityIndex(window.qc.selectors.getPlate());
               });
               $(document).keypress(function (event) {
                   var char = getChar(event).toLowerCase();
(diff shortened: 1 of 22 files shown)


scanomatic/ui_server_data/style/qc_norm.css
```diff
@@ -1,7 +1,7 @@
 /* Data Selection*/
-.SelectedExperiment {
+.selected-experiment {
     stroke: black;
-    stroke-width:2
+    stroke-width: 3;
```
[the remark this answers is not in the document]
local-minimum: ignore: it's an svg property https://css-tricks.com/almanac/properties/s/stroke-width/

scanomatic/ui_server_data/js/qc_normHelper.js
```diff
@@ -287,26 +234,22 @@ function projectSelectionStage(level) {
...
     GetMarkExperiment(path, lockKey, (gData) => {
         if (gData.success === true) {
-            dispatch.reDrawExp(`id${row}_${col}`, mark);
-            const queueCurrent = updateQIndexCoord(qIdxOperations.Current);
-            if (queueCurrent.row == row && queueCurrent.col == col) {
-                setExperimentByQidx(qIdxOperations.Next);
-            } else {
-                setExperimentByQidx(qIdxOperations.Current);
-            }
+            window.drawPlateExperiment(`id${row}_${col}`, mark);
```
skymandr: Doesn't look exactly the same, but I don't see anything wrong either.

scanomatic/ui_server/templates/qc_norm.html
```diff
@@ -340,12 +329,7 @@ <h2 style="float: left">
                   </div>
                   <div class="row">
                       <div class="col-md-12">
-                        <span class="mark-experiments-toggle">
-                          Mark experiments
-                          <input id="ckMarkExperiments" type="checkbox" data-toggle="toggle" data-size="mini"/>
```
skymandr: Perhaps a screenshot to show the difference in the UI after removing things?

scanomatic/ui_server_data/js/src/qc/DrawCurvesIntegration.js
```diff
@@ -25,6 +25,15 @@ export default class DrawCurvesIntegration {
             $(selector).hide();
             return;
         }
+
+        const well = $(`#id${focus.row}_${focus.col}`)[0];
+        if (!well) return;
+
+        // This is horrible, but seems only way to access all needed data
+        // in the legacy code.
```
skymandr: Please update comment with instructions for which legacy code this pertains too (which file). Please also add issue and consider referencing it in the comment.

scanomatic/ui_server/templates/qc_norm.html
local-minimum: screenshot from 2018-06-18 14-15-29
You'll note the toggle-thing is gone.

scanomatic/ui_server_data/js/qc_normHelper.js
local-minimum: I had to rework a little since I needed to get rid of that `dispatch`-thing.

scanomatic/ui_server_data/js/src/qc/DrawCurvesIntegration.js
local-minimum: Note that the comment is not part of this PR, the block just moved. I can answer your question though, when we have the phenotypes shown in the heatmap in the store is when we will remove it. There's a task for that already on the board.

tests/system/conftest.py
```diff
@@ -43,8 +43,8 @@ def is_responsive(url):
         docker_services.port_for('scanomatic-frontend', 5000),
     )
     docker_services.wait_until_responsive(
-        timeout=30, pause=0.1,
-        check=lambda: is_responsive(url + '/fixtures')
+        timeout=360, pause=0.1,
+        check=lambda: is_responsive(url + '/api/status/jobs')
```
gdetrez: Maybe revert the timeout

PR #409, opened 2018-07-06
Qindex input box

skymandr:
Replaced panel showing current qindex with an input box, and re-purposed the reset-button as a set-button.

Since the "hot keys" for switching plates interfered with using the input in a natural way, I removed those.

qindex_input


skymandr:
> @gdetrez : Is the state updated if you use the little arrows ?

No, it is not. I'm not sure that's possible, since it isn't a react component. It would be fairly easy to remove them, is that preferable?

PR #412, opened 2018-08-03
Qc phenotypes store extension

local-minimum:
__PARTIAL FEATURE__

This builds upon #411 and has those changes in it too so probably wait until that one is merged.

Adds the ability to store and get the phenotypes as well as the qc-marks from the store.
Unfortunately there's need for more things to move to the store before it can be integrated, but rather keep PR smaller.


scanomatic/ui_server_data/js/src/qc/actions.js
```diff
@@ -102,3 +152,67 @@ export function retrievePlateCurves() : ThunkAction {
...
+
+export function retrievePhenotypesNeededInGraph(plate: number) : ThunkAction {
+    return (dispatch, getState) => {
+        const state = getState();
+        const project = getProject(state);
+        if (project == null) {
+            throw new Error('Cannot retrieve phenotype if project not set');
+        }
+        const currentPlate = getPlate(state);
+        if (plate !== currentPlate) return Promise.resolve();
```
gdetrez: This looks weird. Is there a reason the code would try to get data for a plate that is not the current plate?

scanomatic/ui_server_data/js/src/qc/actions.js
```diff
@@ -102,3 +152,67 @@ export function retrievePlateCurves() : ThunkAction {
...
+
+        const promises = ['GenerationTime', 'GenerationTimeWhen', 'ExperimentGrowthYield']
+            .filter(phenotype => !hasPhenotypeData(state, phenotype))
+            .map(phenotype => getPhenotypeData(project, plate, phenotype).then((data) => {
+                dispatch(setPlatePhenotypeData(plate, phenotype, data.phenotypes));
+                dispatch(setPhenotypeQCMarks(
```
gdetrez: This look a bit like an anti-pattern: actions that are dispatched together and from the same data. I think it's better to create a higher-level action (e.g. "receive phenotype data") and move the logic in the reducer(s).

scanomatic/ui_server_data/js/src/qc/state.js
```diff
@@ -11,9 +11,65 @@ export type QualityIndexInfo = {
...
+    | "GenerationTimeWhen"
+    | "InitialValue";
+
 export type Settings = {
     +project?: string,
-    +phenotype?: string,
+    +phenotype?: Phenotype,
+};
+
+export type PhenotypeDataCollection = {
```
gdetrez: Wouldn't this be better as a `Map` from `Phenotype` to `PlateValueArray`?

scanomatic/ui_server_data/js/src/qc/state.js
```diff
@@ -11,9 +11,65 @@ export type QualityIndexInfo = {
...
+};
+
+export type QCMarks = {
+    +badData?: PlateCoordinatesArray,
+    +empty?: PlateCoordinatesArray,
+    +noGrowth?: PlateCoordinatesArray,
+    +undecidedProblem?: PlateCoordinatesArray,
+};
+
+export type QCMarksCollection = {
```
gdetrez: Here too (a `Map`).

scanomatic/ui_server_data/js/src/qc/actions.js
local-minimum: Yes, I intended this thunk to be called upon resolving the other thunk:

```javascript
const plate = 42;
retrievePlatePhenotype(plate)
    .then(() => {
        retrievePhenotypesNeededInGraph(plate);
    });
```
Before the call to this function the user could have changed plate in the ui.

scanomatic/ui_server_data/js/src/qc/actions.js
local-minimum: Will fix, forgot to revert to that pattern. Had a version of my code halfways through when marks wheren't always set and then it seemed more reasonable.

PR #416, opened 2018-08-08
QC-Marking individual curves [builds upon #412]

local-minimum:
This makes it possible to mark positions without waiting for the request to the api to go through. However the pending marked position is locked/dirty until the request goes through.

It contains the changes in #412, so probably process that one first.


scanomatic/ui_server_data/js/src/api/setCurveQCMark.js
```diff
@@ -0,0 +1,52 @@
...
+    project: string,
+    plate: number,
+    row: number,
+    col: number,
+    mark: QCMarkType,
+    key: string,
+) : Promise<> {
+    const uri = `/api/results/curve_mark/set/${mark}/${plate}/${row}/${col}/${project}?lock_key=${key}`;
+    return API.postJSON(uri, {})
+        .then((r) => {
```
skymandr: Check if this is needed.

local-minimum:
The comment sent me down a two days rabbit hole so while I think it is true that it isn't needed I think its better to deal with that some other time.

